PR #51, opened 2016-02-02
Add delete animation

naitian:
Add animation that makes it more obvious when you're deleting an entry.


sallar:
Thanks for this :+1:

Issue #28, opened 2016-01-29
Password generator


perry-mitchell:
Implement a password generator

PR #57, opened 2016-02-04
Generate password and new delete alert

naitian:
Uses the [password-generator library](https://github.com/bermi/password-generator) to generate 8-character long memorable passwords. These are not the _most_ secure, though. 

Possible improvements include adding settings for password generation, allowing password restrictions (need special characters, numbers, upper/lower case, etc.).

This would (at least partially) address #28.

Also changed the delete alert box to a sweetalert to make the style fit in more with the rest of the interface.


perry-mitchell:
This looks really good!

If you fix the merge conflict that's just arisen I'll merge it in. I will probably make the default-generated password a bit stronger (more characters and combinations) until we have a more custom approach.

naitian:
Fixed the merge conflict. I also changed the password settings to 12 characters and include both letters and numbers.

naitian:
The test failed and I'm not sure why... Can someone enlighten me?

perry-mitchell:
@naitian I re-ran the single failed test - Travis is buggy imo. No reason for the failure. I'll merge this.

Issue #60, opened 2016-02-05
Entries searching


mpourismaiel:
I have two entries 'Buttercup' and 'Some other website with a cup' and searching 'cup' hides both of them. Also searching any other phrase does the same thing.

perry-mitchell:
Thanks @mpourismaiel - That seems to be the case, we'll need to fix the search.

Thinking it might be wise to have it search the entire archive rather than just the current group.

sallar:
Closed by #92

PR #64, opened 2016-02-06
Fix a bug in addEntries function that was breaking the search, also implemented fuzzysearch

mpourismaiel:
size() function is a prototype of Backbone.Collection and the property passed to addEntries was not an instance of Backbone.Collection, although it looked like it. This fixes issue #60.

Other than that, implemented fuzzy search which displays better results, faster.


sallar:
Thanks for this @mpourismaiel, will check it out ASAP. :+1:

perry-mitchell:
Looks good, let's get it in there :)

PR #67, opened 2016-02-07
Removed folder actions from Trash, fixed confirm dialog

sallar:


source/resources/js/app/models/group.js
```diff
@@ -13,7 +13,18 @@ export default Backbone.Model.extend({
...
     },
+
     defaults: {
         'title': 'Untitled'
+    },
+
+    isTrash: function() {
+        let attr = this.get("attributes");
+
+        if (attr && "bc_group_role" in attr && attr.bc_group_role === "trash") {
```
perry-mitchell: This would probably be best accomplished by using the `isTrash` method on a group - But I understand this is on the front-end. Could we have `bc_group_role` hardcoded in the config somewhere, like in the core project? https://github.com/buttercup-pw/buttercup-core/blob/master/source/classes/ManagedGroup.js#L257

sallar: I understand this is not pretty, but don't know what would be test best way to handle it. Any suggestions?

PR #71, opened 2016-02-10
Added password strength check (colors)

naitian:
The password input box will display colors ranging from red to green depending on how strong the password is, using the zxcvbn library. This will at least partially address #1. The colors could use some tweaking.


sallar:
Thanks @naitian, I’ll check it out soon :) Cheers

PR #72, opened 2016-02-12
Add confirmation for password generation

naitian:
To avoid accidentally erasing passwords from generating a new password, a confirmation dialog box new pops up if the password box is already populating when generating a new password.

### DIFF
--- a/source/resources/js/app/tools/generate.js
+++ b/source/resources/js/app/tools/generate.js
@@ -38,7 +38,6 @@ export function generatePassword(length) {
     const randomLength = length || Math.floor(Math.random() * (maxLength - minLength)) + minLength;
     while (!isStrongEnough(password)) {
         password = passgen(randomLength, false, /[\w\d\?\-]/);
-        console.log(password);
     }
     return password;
 }
--- a/source/resources/js/app/tools/generate.js
+++ b/source/resources/js/app/tools/generate.js
@@ -38,6 +38,7 @@ export function generatePassword(length) {
     const randomLength = length || Math.floor(Math.random() * (maxLength - minLength)) + minLength;
     while (!isStrongEnough(password)) {
         password = passgen(randomLength, false, /[\w\d\?\-]/);
+        console.log(password);
     }
     return password;
 }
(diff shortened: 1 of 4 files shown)


source/resources/js/app/tools/generate.js
```diff
@@ -38,6 +38,7 @@ export function generatePassword(length) {
     const randomLength = length || Math.floor(Math.random() * (maxLength - minLength)) + minLength;
     while (!isStrongEnough(password)) {
         password = passgen(randomLength, false, /[\w\d\?\-]/);
+        console.log(password);
```
sallar: Extra `console.log` here?

source/resources/js/app/tools/strength.js
```diff
@@ -0,0 +1,26 @@
+"use strict";
```
sallar: I think it would be a good idea to merge from master, since this file has already been added in the other PR.

source/resources/js/app/views/entry.js
```diff
@@ -82,8 +88,23 @@ export default Backbone.View.extend({
...
+            $field.first().val(generatePassword());
+        }
+        else {
+            confirmDialog(
+                "Generate a new password?",
+                "This would replace your current password.",
+                "Generate",
+                (confirm) => {
+                    if (confirm === true) {
+                        console.log("confirmed");
```
sallar: `console.log`

sallar:
Hey, thanks for the PR. I added some comments on the code.

naitian:
Thank you for pointing out all of the unnecessary console.logs. I totally forgot to take them out. I'll fix that as soon as I can. Cheers!

source/resources/js/app/views/entry.js
```diff
@@ -164,7 +178,8 @@ export default Backbone.View.extend({
     removeEntry: function () {
         confirmDialog(
             `Delete ${this.model.get("title")}?`,
-            `Are you sure you want to delete this entry?`,
+            `Are you sure you want to delete this entry? This cannot be undone.`,
```
sallar: I had removed `This cannot be undone part` since the entries will be moved to trash now.

PR #155, opened 2017-01-20
Change minimum length of generated passwords to 10 chars

perry-mitchell:
Updating the length so it can be a minimum of 10 characters makes for some faster-typed passwords:

image

This is also inline with [NIST's recommended minimum length](https://nakedsecurity.sophos.com/2016/08/18/nists-new-password-rules-what-you-need-to-know/) of 8 characters.

The generated passwords are still quite secure, and can be more practical than the 20 char ones. This update just allows for more flexibility.

### DIFF
--- a/src/renderer/components/generator.js
+++ b/src/renderer/components/generator.js
@@ -103,7 +103,7 @@ class Generator extends Component {
             <input
               type="range"
               value={this.state.length}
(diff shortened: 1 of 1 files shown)


sallar:
Thanks! 🙌🏻

PR #166, opened 2017-03-09
Move Groups to Root

### LINKED ISSUE #147: Cannot move group to root level
While dragging a group and trying to place it on the root level, an indicator line is shown like it would work, but dropping it has no affect (group stays where it was).

sallar:
This PR adds:

+ moving groups to other groups by right clicking on them
+ moving groups to root by dragging to gaps between groups or by right clicking
+ font smoothing in the UI

Fixes #147

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
-os:
-  - osx
+#os:
+#  - osx
 
 language: node_js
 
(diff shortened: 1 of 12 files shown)


perry-mitchell:
Oh yeah!

PR #186, opened 2017-03-19
Add crash reporter to main process

### LINKED ISSUE #178: Implement Electron’s Crash Reporter
It would be nice to have the `crash-reporter` installed in Buttercup so we can have some usage logs. [More Info](https://github.com/electron/electron/blob/master/docs/api/crash-reporter.md)

sallar:
Closes #178

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -20,6 +20,22 @@ if (process.env.NODE_ENV === 'development') {
   });
 }
 
+// Crash reporter for alpha and beta releases
+// After we come out of beta, we should be rolling our own
+// Crash reporter server using Mozilla Socorro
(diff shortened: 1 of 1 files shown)


src/main/app.js
```diff
@@ -20,6 +20,22 @@ if (process.env.NODE_ENV === 'development') {
   });
 }
 
+// Crash reporter for alpha and beta releases
+// After we come out of beta, we should be rolling our own
+// Crash reporter server using Mozilla Socorro
+// https://github.com/mozilla/socorro
+// This process is fail-safe. Even if the URL stops working
+// The app has already crashed. lol.
```
perry-mitchell: 😂

PR #188, opened 2017-03-24
WIP: File Manager

sallar:


src/renderer/components/icon.js
```diff
@@ -0,0 +1,25 @@
...
+
+  if (!(className in styles)) {
+    className = 'icon-document';
+  }
+
+  return (
+    <i
+      className={cx(styles.icon, styles[className])}
+      style={{ fontSize: `${size}px` }}
+      ></i>
```
perry-mitchell: Can't these just be `<i />`?

src/renderer/components/recent-files.js
```diff
@@ -1,34 +1,54 @@
...
 import styles from '../styles/recent-files';
 import { parsePath } from '../system/utils';
 import { showContextMenu } from '../system/menu';
+import EmptyView from './empty-view';
 
-const File = ({filename, onClick, onRemoveClick}) => {
-  const { base, dir } = parsePath(filename);
+function getIcon(type) {
+  switch (type) {
+    case archiveTypes.OWNCLOUD:
```
perry-mitchell: Should we maybe keep these icons consistent? https://github.com/buttercup/buttercup-browser-extension/tree/master/resources/providers

src/renderer/containers/archive/entry.js
```diff
@@ -1,11 +1,12 @@
 import { connect } from 'react-redux';
 import { initialize, isDirty } from 'redux-form';
 import Entry from '../../components/archive/entry';
-import { getCurrentEntry, updateEntry, newEntry, deleteEntry, changeMode } from '../../redux/modules/entries';
+import { getCurrentEntry } from '../../../shared/selectors';
+import { updateEntry, newEntry, deleteEntry, changeMode } from '../../../shared/actions/entries';
```
perry-mitchell: Should we start to already look at standardising these paths? If we're using webpack we can just alias the root source directory or `renderer` directory so we can use absolute paths.

src/shared/buttercup/types.js
```diff
@@ -0,0 +1,6 @@
+export const archiveTypes = {
```
perry-mitchell: Should this perhaps be uppercase `ArchiveTypes`? Just a bit weird seeing it as camel 🙃

PR #218, opened 2017-04-16
Upgrade buttercup-core-web

### LINKED ISSUE #202: Deleted entries and custom fields re-appearing after a while (WebDAV)
Hi, I have a strange issue, using Buttercup 0.11.1-alpha with a WebDAV-Server - after a while and various restarts, old and already deleted entries, folders and/or custom fields are re-appearing in the app. So, I have e.g. four custom fields for an entry which were deleted either one or two days ago and which have been away for at least a day - but now they are back. I also have some folders in the "Trash", although I have already clicked "Empty trash" at least twice since deleting those entries ....
I am pretty much the only person actively using this WebDAV server at the moment, so I have no idea where those "ghost" entries are coming from... any idea? Thanks!

### LINKED ISSUE #214: [UI] Update current archive in GUI after merging during save
When saving, a merge will occur in certain situations and the contents of the archive might change. Groups and their contents may be completely different, so a full update of the UI might be necessary.

sallar:
Thanks to @perry-mitchell, this PR:
+ Fixes #202 
+ Fixes #214 

Related PR: https://github.com/buttercup/buttercup-core/pull/149

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "babili-webpack-plugin": "^0.0.8",
     "buttercup-importer": "~0.7.1",
     "buttercup-ui": "^0.5.1",
-    "buttercup-web": "^0.25.2",
+    "buttercup-web": "~0.25.2",
     "classnames": "^2.2.5",
     "concurrently": "^2.1.0",
     "cross-env": "^1.0.8",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -118,7 +118,7 @@
     "babili-webpack-plugin": "^0.0.8",
     "buttercup-importer": "~0.7.1",
     "buttercup-ui": "^0.5.1",
-    "buttercup-web": "~0.23.0",
+    "buttercup-web": "^0.25.2",
```
perry-mitchell: We should probably leave web patch locked `~0.25.2`

sallar: Ah yeah @perry-mitchell

PR #226, opened 2017-04-24
Show untitled for groups that don't have any titles

### LINKED ISSUE #190: It's not possible to select a nameless group
I ended up with a nameless group during import from KeePass 1.x to 2.x, and I didn't notice that. Buttercup then happily imported the nameless group from the KeePass 2.x -file, but the UI makes it impossible to select it.

image

Moving everything else out of the parent group and deleting it was a reasonable workaround for me, but it is not the best possible option. I think right-clicking on the folder icon should also be made to work.

sallar:
Fixes #190

### DIFF
--- a/src/renderer/components/tree-view/tree-label.js
+++ b/src/renderer/components/tree-view/tree-label.js
@@ -56,7 +56,7 @@ class TreeLabel extends Component {
               this._input = c;
             }}
             />
-        ) : title}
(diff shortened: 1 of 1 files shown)


perry-mitchell:
This probably doesn't need a release I guess.. What do you think @sallar ?

sallar:
@perry-mitchell It will go out with next release. We don't need to release right away

PR #256, opened 2017-05-17
Fix password dialog cancel error

sallar:


perry-mitchell:
@sallar One of the checks failed sir..

sallar:
@perry-mitchell this too :)

src/shared/actions/archives.js
```diff
@@ -18,25 +18,27 @@ export const removeArchive = createAction(ARCHIVES_REMOVE);
...
-          ? err.message || unknownMessage
-          : unknownMessage
-      );
-    })
-  );
+  try {
+    const archive = await showPasswordDialog(
+      password => loadWorkspace(payload, password).catch(err => {
+        const unknownMessage = 'An unknown error has occurred';
+        return Promise.reject(
```
perry-mitchell: Why reject? You could just throw another error as you're within a Promise anyhow - it'd work the same way and be a little neater (seeing as you're using await). What about:

```javascript
if (isError(err)) {
  throw err;
} else {
  throw new Error(unknownMessage);
}
```

sallar: @perry-mitchell this one is for the password dialog component (swal) to handle and it wants a promise. the await thingie is for the result of that :D

PR #286, opened 2017-06-09
Fix deep find method

### LINKED ISSUE #283: Impossible to create New Group
Hello,

I just installed Buttercup for the first time. I wanted to create a new group, but when I click on the button "+ New  Group" i get an "Unhandled Error".

Here is a screenshot:
image

sallar:
Fixes #283

### DIFF
--- a/src/shared/utils/collection.js
+++ b/src/shared/utils/collection.js
@@ -73,6 +73,10 @@ export function deepAdd(list, id, key, newItem) {
 }
 
 export function deepFindById(list, id, key) {
+  if (!list || !Array.isArray(list)) {
+    return null;
+  }
+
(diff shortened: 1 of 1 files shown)


src/shared/utils/collection.js
```diff
@@ -73,6 +73,10 @@ export function deepAdd(list, id, key, newItem) {
 }
 
 export function deepFindById(list, id, key) {
+  if (!list || !Array.isArray(list)) {
```
perry-mitchell: Wouldn't just `!Array.isArray(list)` work?

sallar: @perry-mitchell you are right sir

PR #296, opened 2017-06-13
Add link to AUR package

perry-mitchell:
Add ref in readme to AUR package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you're using macOS, you can also use **Homebrew Cask** to download and instal
 $ brew cask install buttercup
 ```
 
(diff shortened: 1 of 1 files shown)


Defman21:
AUR package has boot 32 and 64 bit versions.

perry-mitchell:
@Defman21 @sum01 Good call. I don't necessary like the idea of supporting 32bit platforms anymore, but if it's available then fine.

sum01:
@perry-mitchell heh, I can remove the 32-bit stuff from the AUR if you want

perry-mitchell:
> I can remove the 32-bit stuff from the AUR if you want

Perhaps it's more work to remove it than it is to update the readme 😅 - Thanks for the offer. Great that we've got yet another supported platform! 🙏

PR #298, opened 2017-06-18
Change Auto Update System to electron-updater

sallar:


sallar:
@perry-mitchell yes it is backwards compatible. So people that are having the old versions will update to this new version using the old updater (download.buttercup.pw), and new ones will continue getting their updates from new system (github releases).

perry-mitchell:
This is good to go when you're ready @sallar 👍

PR #299, opened 2017-06-18
Change AUR link because of repo rename

sum01:
Only changed the link, but it seems that my Beautify removed some whitespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ If you're using macOS, you can also use **Homebrew Cask** to download and instal
 $ brew cask install buttercup
 ```
 
-Buttercup is available for **macOS (dmg)**, **Windows (exe)** and **Linux (deb, rpm, tarball)** (64bit only). Buttercup is also available for [Arch linux (32/64bit) (AUR)](https://aur.archlinux.org/packages/buttercup/).
+Buttercup is available for **macOS (dmg)**, **Windows (exe)** and **Linux (deb, rpm, tarball)** (64bit only). Buttercup is also available for [Arch linux (32/64bit) (AUR)](https://aur.archlinux.org/packages/buttercup-desktop/).
 
 ## Encryption & Format
 
@@ -55,7 +55,7 @@ If you're interested in developing Buttercup:
 ### Install Dependencies & Run
 
 ``` bash
-$ npm install 
+$ npm install
 $ npm run start
 ```
 
@@ -63,7 +63,7 @@ $ npm run start
 
 ### Install Dependencies
 
-You will need some extra dependencies to build for different platforms on a single platform. Please refer to [this guide](https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build) and install required software for your platform. 
+You will need some extra dependencies to build for different platforms on a single platform. Please refer to [this guide](https://github.com/electron-userland/electron-builder/wiki/Multi-Platform-Build) and install required software for your platform.
 
 ### Building libraries before releasing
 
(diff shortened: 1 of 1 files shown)


sallar:
@sum01 Thanks. but does this mean that `buttercup` package wont be available? I dont think that the package name needs to change also. Since we have the same in `brew` also.

sum01:
@sallar What do you mean unavailable? I just had to remake the repo since there's no way to do namechanges on AUR (I did submit a merge, but those need review from TU's).

sallar:
@sum01 Sorry Im not sure how AUR works :D What name would you use to install from AUR? Cant we leave that as "buttercup" or is it tied to the github repo name?

sum01:
@sallar Well the previous `buttercup` AUR pkgbuild pulls everything from `buttercup` github, and while I can change the actual links to make it continue working, from what I've read the way I went about this is the correct way to do it (aka make new repo, and submit a merge request), because I can't change the actual package name.

But anyways, the AUR works almost exactly the same as the official Arch-linux repo's (meaning package manger-esque functionality), but it's "unofficial" in that literally anyone can upload pkgbuild's. People normally use things called [AUR helpers](https://wiki.archlinux.org/index.php/AUR_helpers) to install from the AUR, although you can manually `wget` and `makepkg` yourself.

sum01:
merge-request Seems the merge went through, which means the old `buttercup` PKGBUILD got deleted, and comments/votes got put into the new one.

perry-mitchell:
Ok.. apologies if I've missed something here. I think we want to keep "buttercup" for the installed package and not "buttercup-desktop" as is mentioned on AUR.

sum01:
> For the sake of consistency, pkgname should match the name of the source tarball of the software: for instance, if the software is in foobar-2.5.tar.gz, use pkgname=foobar.

I simply went by the AUR rules, as stated in [the Arch wiki, under PKGNAME.](https://wiki.archlinux.org/index.php/PKGBUILD#pkgname).

perry-mitchell:
Well regardless, buttercup-desktop is not what we want for the final filenames.. and Sallar, correct me if I'm wrong, but the built source tarball on deployment should still be called buttercup and not buttercup-desktop right?

I understand that there's a standard approach, but we don't have multiple packages being installed - we only provide "buttercup" - the naming of the GitHub repos should not affect the end product.

perry-mitchell:
Just checked the [releases](https://github.com/buttercup/buttercup-desktop/releases/tag/v0.17.2) and it seems that the filenames are just the tag name, at least for the source. @sum01 where would you expect buttercup-desktop to show up? If you mean the downloading of master, that also seems kind of counter intuitive imo, but maybe I don't understand the AUR community so well.

sum01:
Sorry, not really sure I get what the issue is. It's pretty normal to name the AUR pkg after the repo name.

perry-mitchell:
Unless I misunderstand, wouldn't this install the package on an arch linux user's machine as "buttercup-desktop"? If so that's what we're trying to avoid.

All I'm getting at is that we won't be providing another executable really so having "buttercup" as the installed package name would be more fitting. I'm hoping the installed name is not tied to the repository name, as that doesn't make any sense imo.

Does that make any sense? 😅

sum01:
Feel free to upload whatever you want if you disagree, but I see no reason to have the AUR pkg named `buttercup` if the repo is `buttercup-desktop`. It will still appear when people search for buttercup, so finding it is a non-issue.

perry-mitchell:
Ok.. I don't think we completely see eye to eye but I don't feel it's important enough to keep debating the naming right now. I vote we move forward and merge. @sallar ?

sallar:
I agree. Thanks @sum01 lets close and re-open this so the status checks pass.

Issue #282, opened 2017-06-08
Hide menubar in Windows 10


delk0:
Looks like this at the moment in Windows (possibly in linux as well)
![butter](https://user-images.githubusercontent.com/19197011/26943147-e263c332-4c84-11e7-91fe-be68e4624921.PNG)

[here's](https://electron.atom.io/docs/api/browser-window/#winsetmenubarvisibilityvisible-windows-linux) some info that could help you fix this.

Great work guys, loving the app. This was literally the only "flaw" I've managed to find haha.

PR #332, opened 2017-09-12
Issue #282 - Hide menubar in Windows 10

montrueque:


perry-mitchell:
Closes #282.

@sallar Is this how we want it? I guess it should be toggle-able. There are many important functions within the menu that are not accessible otherwise, and we wouldn't want to hinder usability.

src/main/windows.js
```diff
@@ -24,6 +24,8 @@ export function setupWindows() {
       vibrancy: 'ultra-dark'
     });
 
+    win.setMenuBarVisibility(false);
```
perry-mitchell: @sallar Should this not make it hidden always?

sallar: @perry-mitchell if this is set to `false`, the user can still bring the menu up by pressing `alt`. From the docs:

> Sets whether the menu bar should be visible. If the menu bar is auto-hide, users can still bring up the menu bar by pressing the single Alt key.

But I personally thing we should make this toggle-able by having a menu item in View menu for example lables "Auto Hide Menubar" which should only be available in windows.

perry-mitchell: Agreed. Even having it default to shown (first install) and being able to hide it _later_ with alt would be preferable.

sallar:
Thanks @montrueque for the contribution. I took your PR and built on top of it #337. Cheers.

PR #334, opened 2017-09-15
High Priority 1

### LINKED ISSUE #35: Password verification field
I had a typo while setting the password, there is no way to restore it. so adding a verification field might help.

### LINKED ISSUE #219: Closing open archive too early causes loss of data
When adding a new entry, for instance, if the window is closed immediately after creating there is then a high chance of data loss (the latest entry save-request is not completed).

### LINKED ISSUE #221: Add close modal for when closing during an ongoing save process
Add a modal that shows when pressing the close button if a save is ongoing. Sometimes network saves can be slow, and closing too early can kill them.

This fixes #219

### LINKED ISSUE #261: Confirm master password
The master password, when being entered for the first time or changed, should have at least another input to confirm that the typed password is correct. This will help prevent corruption.

### LINKED ISSUE #329: Error when pressing cancel during login
image

Steps:
 1. Open archive
 2. Enter wrong password
 3. Press cancel

sallar:
Fixes #329 
Fixes #219 
Fixes #221 
Fixes #35 
Fixes #261

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -1,4 +1,4 @@
-import { app } from 'electron';
+import { app, ipcMain as ipc } from 'electron';
 import pify from 'pify';
 import log from 'electron-log';
 import { throttle } from 'lodash';
@@ -7,6 +7,7 @@ import configureStore from '../shared/store/configure-store';
 import { setupMenu } from './menu';
 import { getWindowManager } from './lib/window-manager';
 import { loadFile } from './lib/files';
+import { getMainWindow } from './utils/window';
 import { isWindows } from '../shared/utils/platform';
 import { setupActions } from './actions';
 import { setupWindows } from './windows';
@@ -22,6 +23,7 @@ const windowManager = getWindowManager();
 
 let appIsReady = false;
 let initialFile = null;
+let isSaving = false;
 
 // Crash reporter for alpha and beta releases
 // After we come out of beta, we should be rolling our own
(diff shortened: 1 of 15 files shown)


src/shared/actions/archives.js
```diff
@@ -49,7 +49,7 @@ export const unlockArchive = payload => dispatch => {
     password => unlockArchiveInArchiveManager(payload, password)
   ).then(
     archiveId => dispatch(loadArchive(archiveId))
-  );
+  ).catch(() => {});
```
perry-mitchell: Should this dispatch an error message?

src/shared/actions/archives.js
```diff
@@ -65,18 +65,37 @@ export const loadOrUnlockArchive = payload => (dispatch, getState) => {
...
+    .then(firstPassword => showPasswordDialog(password => {
+      if (firstPassword !== password) {
+        return Promise.reject(new Error('Your passwords don\'t match.'));
+      }
+      return addToArchive(password);
+    }, {
+      title: 'Confirm Password'
+    }))
+    .then(dispatchLoad)
+    .catch(() => {});
```
perry-mitchell: This too..

src/shared/actions/archives.js
```diff
@@ -65,18 +65,37 @@ export const loadOrUnlockArchive = payload => (dispatch, getState) => {
...
+        : unknownMessage
+    );
+  });
+
+  // If it's not a new archive,
+  // show the password dialog only once.
+  if (isNew === false) {
+    return showPasswordDialog(addToArchive)
+      .then(dispatchLoad)
+      .catch(() => {});
```
perry-mitchell: And this..

src/main/app.js
```diff
@@ -136,3 +138,26 @@ app.on('activate', () => {
...
+// to prevent data loss
+ipc.on('workspace-save-started', () => {
+  isSaving = true;
+});
+
+ipc.on('workspace-save-finished', () => {
+  isSaving = false;
+});
+
+app.on('before-quit', e => {
```
perry-mitchell: I sent a message about this event and cb. If it blocks the user from doing other save-inducing actions, that would be best. Could you confirm the behaviour here? 😊

src/shared/actions/archives.js
sallar: No, this catch is only for the password dialog "cancel" event.

src/main/app.js
sallar: Yeah, the UI disables everything when this happens. Only thing user can do is wait 😂

src/shared/actions/archives.js
perry-mitchell: What if this fails? `unlockArchiveInArchiveManager(payload, password)`

sallar: That is handled by the dialog (it shows the error message). The dialog is a but weird

Issue #354, opened 2017-10-16
Connected cloud source gets removed


ElPsyCongro:
Hi, I'm using the 0.21.2 version on Windows and when I connect a cloud source and then relaunch Buttercup the cloud source gets removed and I have to add it again, I think this worked well in an earlier version but I do not remember. 
The same version on Mac works well. The cloud that I'm using is DropBox.

danyfu:
Hi, same problem using 0.21.2 version on Windows, and using connect cloud source to Dropbox.

perry-mitchell:
Hmm.. that’s interesting. Do you see any errors throughout the connection or usage of Dropbox?

danyfu:
There is no error, when I use the connection using Dropbox I can work fine and everything is good, but if I close the windows app and reopen it, it opens like there is no connection before and then I need to set the connection to Dropbox again with the credentials.

perry-mitchell:
That’s weird.. Dropbox should save using the same method as the others. Have you by any chance tested a local file or another cloud source?

perry-mitchell:
Yeah it definitely seems to be reproducible. Before:

![image](https://user-images.githubusercontent.com/3869469/31649313-f120a9f6-b31a-11e7-9521-e81c08d08460.png)

After reloading:

![image](https://user-images.githubusercontent.com/3869469/31649360-258cf1ea-b31b-11e7-9b38-bf2742efaa48.png)

sallar:
@ElPsyCongro @danyfu @perry-mitchell I'm working on an update, it will come through tonight hopefully. Could you guys please install that and let us know if it's fixed or not? If not, please re-open this issue (this will auto close by the PR). The version will be `0.21.3`

danyfu:
Thanks @sallar, now its working fine. The Cloud Source persists when close and open the windows app.

👍

ElPsyCongro:
I confirm that this issue was solved. Thanks to the team for the fast attention.

PR #355, opened 2017-10-17
Update the way state is saved, fix menubar autohide

sallar:
This PR fixes:

- Windows data loss #354 
- Auto hide menubar option was being shown to wrong OS #282

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -1,7 +1,7 @@
 import { app, ipcMain as ipc } from 'electron';
 import pify from 'pify';
 import log from 'electron-log';
-import { throttle } from 'lodash';
+import debounce from 'lodash/debounce';
 import jsonStorage from 'electron-json-storage';
 import configureStore from '../shared/store/configure-store';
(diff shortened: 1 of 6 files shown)


src/main/app.js
```diff
@@ -1,7 +1,7 @@
 import { app, ipcMain as ipc } from 'electron';
 import pify from 'pify';
 import log from 'electron-log';
-import { throttle } from 'lodash';
+import debounce from 'lodash/debounce';
```
perry-mitchell: I'd almost recommend [this one](https://www.npmjs.com/package/debounce), so that you wouldn't have to use lodash at all..

perry-mitchell: But if it works it works..

perry-mitchell:
I restarted the travis build too, as it failed..

src/main/app.js
sallar: @perry-mitchell if we didnt use lodash anywhere else I would use this, maybe I'll remove lodash dep in the future as its really heavy

PR #360, opened 2017-10-18
Create a new window on file double click

### LINKED ISSUE #358: Opening an archive file (double-click) does not open UI
Double-clicking an archive file does not open the UI. Buttercup is running, but no window appears. Archive is not added to the interface.

sallar:
Fixes #358

### DIFF
--- a/src/main/lib/files.js
+++ b/src/main/lib/files.js
@@ -88,14 +88,14 @@ export function loadFile(filePath, win, isNew = false) {
   }
 
   // If we have found a window at this point
-  if (win !== null) {
+  if (win) {
     win.webContents.send('load-archive', payload);
     return;
   }
 
   // Otherwise create a new window
-  windowManager.buildWindowOfType('main', win =>
-    win.webContents.send('load-archive', payload)
+  windowManager.buildWindowOfType('main', newWindow =>
+    newWindow.webContents.send('load-archive', payload)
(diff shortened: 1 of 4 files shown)


src/main/lib/files.js
```diff
@@ -86,9 +86,17 @@ export function loadFile(filePath, win, isNew = false) {
   if (!win) {
     win = getMainWindow();
   }
-  if (win) {
+
+  // If we have found a window at this point
+  if (win !== null) {
```
perry-mitchell: What about joining this to the previous if?
```javascript
   if (!win) {
      win = getMainWindow();
    } else {
        win.webContents.send('load-archive', payload);
        return;
    }
```

perry-mitchell: Oh, actually I see why.. it occurs after. Would it be safer to just check truthiness? `if (win) {`?

src/main/lib/files.js
```diff
@@ -86,9 +86,17 @@ export function loadFile(filePath, win, isNew = false) {
...
     win.webContents.send('load-archive', payload);
+    return;
   }
+
+  // Otherwise create a new window
+  windowManager.buildWindowOfType('main', win =>
```
perry-mitchell: Same inner variable.. what about `newWindow`?

PR #365, opened 2017-10-20
Increase update frequency, improve menu items

### LINKED ISSUE #351: Check for updates more often
There should be an option that user can trigger "Check for updates" manually. The app should also check for updates from time to time.

sallar:
Fixes #351

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -1,5 +1,5 @@
 import { app, shell, Menu } from 'electron';
-import { isOSX } from '../shared/utils/platform';
+import { isOSX, isWindows } from '../shared/utils/platform';
 import {
   getCurrentArchiveId,
   getAllArchives,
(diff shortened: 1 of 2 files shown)


src/main/menu.js
```diff
@@ -142,14 +143,21 @@ if (isOSX()) {
   );
 }
 
+// Check for Updates...
+defaultTemplate[isOSX() ? 0 : 4].submenu.splice(isOSX() ? 1 : 0, 0, {
```
perry-mitchell: Woah.. Don't mean to be a buzzkill.. but is there a nicer way to do this? This be crazy hard to read 😅

src/main/menu.js
```diff
@@ -142,14 +143,21 @@ if (isOSX()) {
...
 export function setupMenu(store) {
   const state = store.getState();
   const archives = getAllArchives(state);
   const currentArchiveId = getCurrentArchiveId(state);
-  let condenced = Boolean(getSetting(state, 'condencedSidebar'));
-  let menubarVisible = getSetting(state, 'menubarVisible');
 
   // Default should be safe (always on) in case it isn't set.
+  let menubarVisible = getSetting(state, 'menubarVisible');
   menubarVisible = typeof menubarVisible === 'boolean' ? menubarVisible : true;
```
perry-mitchell: What about:

```javascript
const menubarVisibleSetting = getSetting(state, 'menubarVisible');
const menubarVisible = typeof menubarVisibleSetting === 'boolean' ? menubarVisibleSetting : true;
```

?

PR #367, opened 2017-10-21
Queuing Save Operations

### LINKED ISSUE #363: Error when saving a new entry with multiple empty fields and then close the application
Hello again guys,

Today I faced two bugs in a row, let me explain how I found them:

First of all, I was adding a new entry, after entering a custom field I started adding more but these ones where empty because I thought I was going to need them, but I wasn't, so I just hited the save button just expecting that if the fields were empty, they weren't going to be saved, but just ignored. turns out, when I did that, this error popped up:

buttercup-save-entry-empy-fields

So I just hitted the Ok button, the dialog closed, then I deleted all the blank fields and just leave those I had information, and then saved normally.

Finally, I closed the app because I wasn't going to use it anymore, and then this error popped up:

buttercup-close-error

That's all I did in order to found this errors, if it is helpful, here is my PC specs and buttercup version:

**Buttercup Version**: 0.21.4
**OS**: Windows 10 64-Bit
**CPU**: Intel Core i3-3220 3.30GHz
**RAM**: 16GB
**SSD**: 240 GB

### LINKED ISSUE #366: Process queue for I/O operations
It would be nice if we had a process queue for I/O operations so we can push each save operation to the queue and prevent quitting the app before the queue is empty. This will prevent data loss for all saving operations.

sallar:
Fixes #366 
Fixes #363

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "opencollective": "^1.0.3"
   },
   "devDependencies": {
+    "@buttercup/channel-queue": "^0.2.1",
     "@buttercup/ui": "^0.6.1",
     "@types/node": "^6.0.45",
     "@types/react": "^0.14.41",
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "opencollective": "^1.0.3"
   },
   "devDependencies": {
-    "@buttercup/channel-queue": "^0.2.1",
+    "@buttercup/channel-queue": "~0.2.1",
     "@buttercup/ui": "^0.6.1",
     "@types/node": "^6.0.45",
     "@types/react": "^0.14.41",
--- a/src/renderer/system/queue.js
+++ b/src/renderer/system/queue.js
@@ -1,0 +1,18 @@
+import uuid from 'uuid';
+import { ipcRenderer as ipc } from 'electron';
+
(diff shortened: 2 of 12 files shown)


perry-mitchell:
@sallar This is missing some eslint fixes (see end of appveyor output).

package.json
```diff
@@ -111,6 +111,7 @@
     "opencollective": "^1.0.3"
   },
   "devDependencies": {
+    "@buttercup/channel-queue": "^0.2.1",
```
perry-mitchell: We should probably use `"~0.2.2"` here, for stability's sake.

src/renderer/system/queue.js
```diff
@@ -0,0 +1,18 @@
...
+export function enqueue(channelName, fn, stack) {
+  const id = uuid.v4();
+
+  ipc.send('channel:enqueue', {
+    channelName,
+    stack,
+    id
+  });
+
+  ipc.once(`channel:execute:${id}`, () => {
```
perry-mitchell: I know these are async, but I would feel much safer if this `once` listener was before the `send`.. 😅

sallar:
@perry-mitchell Thanks mate. Changed :)

perry-mitchell:
Good stuff!! 👍

Issue #369, opened 2017-10-22
Show menu bar once auto hide


lalitsom:
Hey I am using this on windows 7 32 bit. I don't know if this is an actual issue or not.
How to show the menu bar again, once you hide it from menu bar->view->auto hide menubar?

sallar:
@lalitsom I just tested it. You are right... it's "supposed" to come back when you hold the `alt` key (like normal windows apps) but it didnt happen for me. Does it come back when you hold alt?

perry-mitchell:
> when you hold the alt key (like normal windows apps)

I'm not sure if holding is how it is normally.. At least I've always pressed it once to show the menu 😅

lalitsom:
No it does not come back when I hold alt.

PR #374, opened 2017-10-26
added keyboard shortcut to toggle visibility of menu bar.

lalitsom:
Fixes Issue #369 
changed label 'auto hide' to 'hide/show'.
added 'ctrl+shift+H' as accelerator.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -157,9 +157,11 @@ export function setupMenu(store) {
   const currentArchiveId = getCurrentArchiveId(state);
 
   // Default should be safe (always on) in case it isn't set.
-  const menubarVisibleSetting = getSetting(state, 'menubarVisible');
-  const menubarVisible =
-    typeof menubarVisibleSetting === 'boolean' ? menubarVisibleSetting : true;
+  const menubarAutoHideSetting = getSetting(state, 'menubarAutoHide');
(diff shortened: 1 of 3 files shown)


src/main/menu.js
```diff
@@ -207,9 +207,10 @@ export function setupMenu(store) {
             ...(isWindows()
               ? [
                   {
-                    label: 'Auto-hide Menubar',
+                    label: 'Hide/Show Menubar',
                     type: 'checkbox',
                     checked: !menubarVisible,
+                    accelerator: 'CmdOrCtrl+Shift+H',
```
perry-mitchell: This is nice, but a couple of things that would be nice to iron out:

 * Why not Alt? Alt is generally accepted (and programmed) in windows to show file/menu bars when they're hidden (as mentioned [here](https://www.computerhope.com/issues/ch000973.htm) for example).
 * If we have to stick to a different accelerator for some reason, is `H` the best key? It could be for **H**ide, sure, but there may be many things in the interface that we want to hide or show. I would suggest `M` personally, for menu bar, but now I'm not sure what other keys we could potentially be stepping on.

What's your opinion @sallar?

lalitsom:
How about using setAutoHideMenuBar(bool) instead of setMenuBarVisibility? This would give the intended default behavior.

PR #378, opened 2017-10-29
Fixed eslint warning

snhasani:
These eslint warnings should be resolved:

image

and also I added `yarn.lock` file. Because I could not install npm packages with Iran's Internet, but Yarn did it!


.eslintrc
```diff
@@ -18,6 +19,7 @@
     "semi": ["error", "always"],
     "space-before-function-paren": ["error", "never"],
     "jsx-quotes": "off",
-    "comma-dangle": "off"
+    "comma-dangle": "off",
```
perry-mitchell: I think this violates the prettier config. I would suggest having comma-dangle as a violation.

@sallar?

sallar: @perry-mitchell I think those already existed in `.eslintrc` before prettier era.. I have added "prettier" preset so all these extra overrides should be removed from here including this one. But I think this changed was not made by @snhasani

perry-mitchell: Ah good point! My mistake

PR #381, opened 2017-10-30
fixed: #379 new group input focus-out handler

### LINKED ISSUE #379: new group directory disappear after focusout
In MacOS app when you want to create a new group if after enter group name you focus out from input to somewhere else, group is not creating and group be disappear.

I think this is a UX issue and user should be able submit a group with focus out by tab, click or whatever it could.

snhasani:
Fixes #379

### DIFF
--- a/src/renderer/components/tree-view/tree-label-edit.js
+++ b/src/renderer/components/tree-view/tree-label-edit.js
@@ -11,6 +11,8 @@ const Input = styled.input`
   color: #222;
 `;
 
+const DEFAULT_TITLE = 'Untitled';
+
 export default class LabelEditor extends Component {
   static propTypes = {
     node: PropTypes.object,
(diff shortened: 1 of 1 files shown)


perry-mitchell:
Thanks @snhasani - I'll let @sallar review. I think it'd be best to hold this for a larger release once merged.

src/renderer/components/tree-view/tree-label-edit.js
```diff
@@ -19,20 +19,25 @@ export default class LabelEditor extends Component {
   };
 
   state = {
-    title: 'Untitled'
+    title: 'Untitle'
```
sallar: Why is **Untitled** changed to **Untitle**?

src/renderer/components/tree-view/tree-label-edit.js
```diff
@@ -19,20 +19,25 @@ export default class LabelEditor extends Component {
...
   };
 
-  handleBlur() {
-    this.props.onDismiss();
+  _handleChange(e) {
```
sallar: Is there a reason why all the handlers now have `_` at the beginning? I know it makes private methods obvious but let's keep the repo's convention for now, what do you think?

src/renderer/components/tree-view/tree-label-edit.js
snhasani: oh, this is a typo! I fix it.

src/renderer/components/tree-view/tree-label-edit.js
snhasani: I think they help to code readability. I am okay with naming without underline, also repo's convention is important anyway. I remove those underlines.

PR #382, opened 2017-10-30
Fix menu bar update frequency

### LINKED ISSUE #373: Import not available
When trying to import into an unlocked archive, the import options are not available:

image

sallar:
Fixes #373

### DIFF
--- a/src/main/actions.js
+++ b/src/main/actions.js
@@ -1,3 +1,4 @@
+import debounce from 'lodash/debounce';
 import { ipcMain as ipc, BrowserWindow } from 'electron';
 import { pushUpdate, updateInstalled } from '../shared/actions/update';
 import { getWindowManager } from './lib/window-manager';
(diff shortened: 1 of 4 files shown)


src/main/actions.js
```diff
@@ -11,6 +12,9 @@ export function setupActions(store) {
   // Clear update notice
   store.dispatch(updateInstalled());
 
+  // Update the menu
+  store.subscribe(debounce(() => setupMenu(store), 500));
```
perry-mitchell: Why debounce menu creation? Isn't 500ms maybe a bit much?

sallar: Its not menu creation. Everytime the store updates (any action regardless of it having anything to do with menu or not) this re-draws the menu which is slightly heavy process. And sometimes this function (subcribed function) gets called 20 times in 100ms. So just wanted to avoid performance issues and lags

perry-mitchell: 👍

Issue #284, opened 2017-06-09
Buttercup in Spanish?


llkiriell:
I want to have the application in Spanish language. I found these files:

![shot_170608_214656](https://user-images.githubusercontent.com/12781734/26959160-4282874e-4c94-11e7-9d6a-f567529a79f2.png)

Exists? How can I change the language?

perry-mitchell:
Thanks @llkiriell - This is blocked by #285. Although those files exist, we do not currently have any translations available. We're looking at adding a translation system so people can contribute and add localisations based upon demand. We should have this soon enough :)

sallar:
Thanks @llkiriell - Those translation files that you mentioned belong to the embedded chrome/electron runtime and not us. But like @perry-mitchell mentioned, we are looking into adding localization.

sallar:
This is done and will be released today. Closing this 🎉

PR #397, opened 2017-11-12
Add localisation support

### LINKED ISSUE #285: Add localisation support
Add support for localisations.

ph1p:
Hey folks,

I've created one type of language support and I've seen the **i18next** branch to late. 
This is why I used another library.

I used react-intl and wrote small spartanic helper functions for main process files.
All languages are located in /locales/*.json.
There is no specific reason why I've used react-intl instead of any other i18n lib.

My main goal was to change as little code as possible and to add new languages easily.

**Menu item**
screen shot 2017-11-12 at 22 13 49

**Some links**
- [Post from the creator of i18next](https://github.com/i18next/react-i18next/issues/303)
- [react-intl](https://github.com/yahoo/react-intl)

**Issues**
#285, #292, #284

Closes #285


sallar:
Thanks @ph1p for the great work. I quite like this. It's ok to use react-intl as far as I can see, since obviously its easy to use the translations in non-react files too.
Before I review the code, could you check why are the tests failing?

ph1p:
@sallar Thank you! I have a problem with the current version of buttercup. I can't run `npm run build:renderer`. This is why I updated all outdated packages (That did not help). 
Webpack freezes at `91% additional asset processing` and node throws **FATAL ERROR: CALL_AND_RETRY_LAST**. 
Your last working commit was `8732150474e9bf6be7cb78faf9bff7f17d9049e2`. 
And the problem is babili (babel-minify). I replaced it with uglify and commented out babel-minify. Now it works, hopefully :)

config/webpack.config.production.js
```diff
@@ -47,7 +70,55 @@ module.exports = merge(baseConfig, {
     new webpack.DefinePlugin({
       'process.env.NODE_ENV': JSON.stringify('production')
     }),
-    new BabiliPlugin(),
+    new UglifyJSPlugin({
```
sallar: Does uglifyJS support ES6 which we are using? I think the reason I used Babili was that it was the only one that supported ES6 out of the box.

config/webpack.config.production.js
```diff
@@ -47,7 +70,55 @@ module.exports = merge(baseConfig, {
...
+          join_vars: true,
+          drop_console: true
+        },
+        output: {
+          comments: false,
+          beautify: false
+        }
+      }
+    }),
+    // new MinifyPlugin({
```
sallar: Can you remove this unused comments?

package.json
```diff
@@ -108,36 +108,37 @@
   },
   "homepage": "https://buttercup.pw",
   "dependencies": {
-    "opencollective": "^1.0.3"
+    "opencollective": "^1.0.3",
+    "react-intl": "^2.4.0"
```
sallar: Can we move this to devDependencies? Since it will be compiled to the final package

src/main/menu.js
```diff
@@ -11,145 +11,148 @@ import { openFile, openFileForImporting, newFile } from './lib/files';
 import { getWindowManager } from './lib/window-manager';
 import { checkForUpdates } from './lib/updater';
 import { getMainWindow } from './utils/window';
+import i18n from '../shared/i18n';
 import pkg from '../../package.json';
 
-const defaultTemplate = [
```
sallar: Is this diff block because you moved this constant to the setup function?

src/renderer/index.js
```diff
@@ -61,10 +63,17 @@ ipc.on('will-quit', () => {
   store.dispatch(setUIState('isExiting', true));
 });
 
+// show message, when locale changed
+ipc.on('locale-changed', (e, payload) => {
+  alert(payload);
```
sallar: Can you use the `showDialog` function in `renderer/system/dialog.js`?

src/shared/i18n/index.js
```diff
@@ -0,0 +1,104 @@
...
+    require(`react-intl/locale-data/${lang.code}`)
+  )
+);
+
+// add all locale files to an array for electron
+const getAllTranslationsAsObject = () => {
+  const translations = [];
+
+  config.availableLanguages.forEach(lang => {
+    translations[lang.code] = require(`../../../locales/${lang.code}`);
```
sallar: I think this wont work when compiled. Since the paths will change. in `menu.js` I do this for resolving: ```path.resolve(__dirname, `../resources/icons/${item.icon}.png`)``` and I think this should be similar.

src/renderer/index.js
sallar: Also, why does the user need to relaunch the program? wont react-intl change the texts in place?

config/webpack.config.production.js
ph1p: Jep. Documentation says, that i can use the `ecma` option (https://github.com/webpack-contrib/uglifyjs-webpack-plugin#uglifyoptions)

config/webpack.config.production.js
ph1p: Sure

package.json
ph1p: Jep

src/main/menu.js
ph1p: Yes, because I need the `store` variable for all menu points.

src/shared/i18n/index.js
sallar: Also can we use `.map` here? 

``` js
const translations = config.availableLanguages.map(lang => require(...));
```

src/renderer/index.js
ph1p: Jep

ph1p: Because I can't find a solution to update the whole electron menu.

sallar: @ph1p We do that when a new archive is added. For example I call it everytime the store gets updated: `store.subscribe(debounce(() => setupMenu(store), 500));`. So this should also update the language already if everything's setup correctly

ph1p: Ok cool. I'll try it :)

locales/de.json
```diff
@@ -0,0 +1,21 @@
...
+  "components.add-archive-button.add_archive": "Archiv hinzufügen",
+  "components.file-manager.index.cancel": "Abbrechen",
+  "components.file-manager.index.go_back": "Zurück",
+  "components.file-manager.index.new_archive": "Neues Archiv",
+  "components.file-manager.index.open_in_buttercup": "Mit Buttercup öffnen",
+  "system.dialog.title": "Master Passwort",
+  "system.dialog.confirm": "Bestätigen",
+  "system.dialog.nevermind": "Abbrechen",
+  "system.dialog.password": "Passwort",
+  "main.menu.language": "Sprache",
```
perry-mitchell: I really like the nested style, though I think some components should be named differently. For instance, `main.menu` alludes to there being a `main` category with a `menu` subcategory, which I don't think there is. Should this be `main-menu` perhaps? Or even `menu.main`?

locales/en.json
```diff
@@ -0,0 +1,21 @@
...
+  "components.add-archive-button.add_archive": "Add Archive",
+  "components.file-manager.index.cancel": "Cancel",
+  "components.file-manager.index.go_back": "Go Back",
+  "components.file-manager.index.new_archive": "New Archive",
+  "components.file-manager.index.open_in_buttercup": "Open in Buttercup",
+  "system.dialog.title": "Master Password",
+  "system.dialog.confirm": "Confirm",
+  "system.dialog.nevermind": "Nevermind",
+  "system.dialog.password": "Password",
+  "main.menu.language": "Language",
```
perry-mitchell: And here as well..

locales/es.json
```diff
@@ -0,0 +1,21 @@
...
+  "components.add-archive-button.add_archive": "Agregar archivo",
+  "components.file-manager.index.cancel": "Cancelar",
+  "components.file-manager.index.go_back": "Volver atrás",
+  "components.file-manager.index.new_archive": "Nuevo archivo",
+  "components.file-manager.index.open_in_buttercup": "Abrir en Buttercup",
+  "system.dialog.title": "Contraseña maestra",
+  "system.dialog.confirm": "Confirmar",
+  "system.dialog.nevermind": "No importa",
+  "system.dialog.password": "Contraseña",
+  "main.menu.language": "Idioma",
```
perry-mitchell: And here..

perry-mitchell:
Also, this shouldn't be merged until we've had a review of the Spanish and German content 😊

ph1p:
I'm german, so I reviewed the german part by myself ^^.  
I only added spanish language, because of ticket #284. 
A girlfriend of mine speaks spanish, she will translate it.
These language files are only examples. I'll write a little explanation about the naming and update all of them.
---
Thank you for reviewing my code. I give my best to solve all problems.

src/shared/i18n/index.js
ph1p: It will compile. I've made the forEach, because if you dynamically require content, webpack typically ignores it. With this little trick I require all translations and load them into the object. So when I add a language to the config object, the translations are automatically added.
Anyways, I have updated my code and decided to import the languages by hand.

ph1p:
I've updated my code and have decided to hold the naming of language ids easier. 
The id describes or is itself the translated word written in lower kebab-case.
I think it's better, because it's shorter and easier to use with duplicate words.

src/renderer/index.js
ph1p: Yeah, It works!

perry-mitchell:
@ph1p That's awesome, and we appreciate it. As a matter of habit I would personally like all translations (first time) verified by at least 1 other person - it's not a matter of trust but more of habit. I can probably have that done today anyhow 😊

perry-mitchell:
Also @sallar - Seeing as this PR adds language **detection**, would it be best to pre-release this to someone running under German/Spanish locale by default?

src/main/menu.js
```diff
@@ -204,6 +215,28 @@ export function setupMenu(store) {
...
               }
             },
+            { type: 'separator' },
+            // Language menu point
+            {
+              label: i18n.formatMessage({
+                id: 'language'
+              }),
+              submenu: i18n.getConfig('availableLanguages').map(lang => ({
+                label: lang.name,
```
perry-mitchell: Is this showing lowercase languages in the menu? Perhaps we could install `ucfirst` to transform the first char to uppercase? What do you think @sallar?

perry-mitchell: Nevermind, I saw that they're capitalised in a later file.

perry-mitchell:
I've had the German verified. The Spanish may get a PR later today, but I'm happy with it being merged. Thanks @ph1p !

ph1p:
> @ph1p That's awesome, and we appreciate it. As a matter of habit I would personally like all translations (first time) verified by at least 1 other person - it's not a matter of trust but more of habit. I can probably have that done today anyhow 😊

Oh and thanks again for the hard work! This was a massive task, and we're just being extra careful with it. Because of this work, other translations will be much easier 🎉

Thanks again (: I totally understand that.

> Also, @sallar + @ph1p - Are we able to verify what happens if a property in one language is missing? I was hoping that it might default to English, but I can't see yet from the code what the behaviour might be if a translation is not present in one locale..

I'll implement a fallback solution.

perry-mitchell:
> I'll implement a fallback solution.

Is that needed? I was just wondering what the behaviour of react-intl was when say a translation exists in English but not in Spanish - Does it fall back automatically?

ph1p:
> Does it fall back automatically?

Unfortunately not :/ There is an extra prop for this called `defaultMessage`. I think I'm ready to push it in about an hour.

perry-mitchell:
@sallar @ph1p Seeing as this is already functional, do you think we could do without the fallback? Perhaps @ph1p could add it in a separate pull request?

I'd love to see this get merged.. I have a feeling that this will open the flood gates on translations 😍

ph1p:
I've added a full translation @perry-mitchell + some bug fixes.

perry-mitchell:
> I can't find the word Trash

Ah.. well this is more complex. `General` and `Trash` are groups - content within the Archive data. These are created as boilerplate items in `buttercup-core`. I agree that these should be translated - I can add an option to the core for this, but it should be done later imo (not part of this PR).

ph1p:
@perry-mitchell Ah ok and take look at password generator, this must also be dynamic.

perry-mitchell:
> password generator, this must also be dynamic

Agreed.. I think we need to figure out a nice way to pass translations around. I guess the generator, being a separate component, should have it's own internationalisation, but should take the current language from the host application (ie. the desktop application in this case).

perry-mitchell:
Hmm.. appveyor is still dying. Issue seems to be with `npm`. @ph1p would you mind trying to add `npm cache verify` to the appveyor build script? Before `npm install`?

ph1p:
Ok guys ...  I've tried several things, but I don't know why travis and veyor failing. There're stopping at `ava && npm run lint`. I've tested it on my own and can't reproduce it.

sallar:
@perry-mitchell @ph1p I'll have a crack at the app veyor thing to see whats dying.

sallar:
@ph1p Im going to do some commits to your repo if you dont mind. including reverting some of the package.json upgrades. It's too many changes for this PR and maybe its one of them that breaks the whole thing.

sallar:
@perry-mitchell @ph1p done!

ph1p:
Thanks @sallar. This was a stupid idea of mine, to push all code into the master branch 🤗. Next time I’ll create an extra feature branch and another pull request.

perry-mitchell:
> This was a stupid idea of mine

No such thing! This was a huge chunk of work. We obviously need to update packages anyway.. 😁

ph1p:
Ok I've one last commit with some bugfixes and translations (no dependency updates). In about 20 minutes I'll push it.

sallar:
Great work @ph1p. I'm going to merge this :)

Issue #371, opened 2017-10-22
Unhandled Error on saving entry with empty custom label name


lalitsom:
![ggg](https://user-images.githubusercontent.com/23455413/31867456-e650df0e-b7ad-11e7-8ad2-95552955df91.JPG)

If user click on Add new field and then forget to enter label name, the save button will give this not so user friendly error message.
Maybe you could give a nice message that "Please enter label name" or maybe you can give some default label name.

snhasani:
@sallar you can assign this to me.

I suggest we can ignore empty custom fields and we save just dirty fields. let me know what you think about this.

perry-mitchell:
All fields require unique names, so we do require users to actually enter names before saving. Perhaps an error message is inappropriate, but saving should be prevented and the user notified somehow that they should enter the field name.

snhasani:
@perry-mitchell I totally agree with you, but my suggest specifically is related to the custom field. Because I think its nature is optional and if user add a custom field but didn't fill it, we can just ignore it when we want submit the form. But we could have some validation rule for its value.

perry-mitchell:
What if a user painfully types out a large amount of text and we delete it when saving because it didn’t have a label?

I would still suggest blocking saving if any part of the form is invalid. Saving while there is invalid data, even by skipping fields, is misleading I feel.

I think the whole red-highlight kind of style when submitting incomplete forms is a nice way to accomplish this, and isn’t as in-your-face as an error dialog.

snhasani:
@perry-mitchell I did not mean we should just have a validation in submit level. I agree with you and I think we must be have validation for each field on field level when blur event.

My mean from ignoring custom field it was when user did not filed any of field inputs. We can just ignore that field when both of them (name and value) was empty.

perry-mitchell:
> We can just ignore that field when both of them (name and value) was empty.

Ah, I understand what you meant.. Yes that definitely sounds correct to me also. I would go as far to say that we could delete that _row_ if it is entirely empty.

snhasani:
So, based on my knowledge I must hit these:

1. Field level validation with good messaging (instant)
2. Submit level normalize and validation with good messaging

And for implement these items I think current design is not good for and we need discuss about this.

For example on field validation we don't have any space between fields row and we can not show the error message under or top of them.

And for custom fields if we want have field validation we need to show inputs with stack view.

cc @sallar @perry-mitchell

ph1p:
I've fixed this error in my PR #400

PR #400, opened 2017-11-15
Handling form validation, fix some i18n bugs and add missing translations

ph1p:
Buttercup shows error popups on saving empty meta data objects (custom fields). This PR should fix it.
Maybe this is a part of buttercup-core.

**Reproduce**
- Create or edit entry
- add empty custom field(s)
- click save

**Issues**: #363, #371


perry-mitchell:
Thanks for the PR @ph1p - Unfortunately I don't know if this is the best way to accomplish this fix. It's absolutely a problem, but forcibly filtering out "invalid" or empty meta values is perhaps a bit heavy handed and dangerous. It would be best to have some kind of **form validation** that is executed when clicking save. Errors could be pointed out and saving prevented, if there are issues.

ph1p:
You're right it's a little bit quick an dirty. I'm already on it!
I'm renaming this PR to "Handling Form Validation".

perry-mitchell:
@ph1p Remember to not add to much to the same PR.. Every pull request should target perhaps 1 issue and not try to do too much else. I wouldn't personally add any more translation content because we have to re-verify all of that.

Keep up the awesome work! ☺

ph1p:
My mistake. While I was coding, I found these errors/untranslated words and fixed them immediately. Next time I'll write them down and split them into different PRs. This PR is ready so far or should I split it now and create new PRs?

src/renderer/components/archive/entry-view.js
```diff
@@ -29,14 +29,19 @@ const EntryView = ({ entry, intl }) => (
...
-          <div className={formRow} key={meta.key}>
-            <div className={labelWrapper}>{meta.key}</div>
-            <div className={inputWrapper}>
-              <Copyable>{meta.value}</Copyable>
-            </div>
-          </div>
-        ))}
+        {entry.meta.map(
+          meta =>
+            Object.keys(meta).length ? (
```
perry-mitchell: You could use jsx control statements here:

```javascript
<If condition={Object.keys(meta).length > 0}>
  <!-- content -->
</If>
```

perry-mitchell: This might be more tidy than the ternary operator.

src/renderer/components/tree-view/tree-label-edit.js
```diff
@@ -76,3 +82,12 @@ export default class LabelEditor extends Component {
     );
   }
 }
+
+LabelEditor.propTypes = {
+  node: PropTypes.object,
```
perry-mitchell: Node is used without checks so should be `isRequired` here.

perry-mitchell: I know that you didn't invent the code of course, but if we're changing it then it's a good time to update it 😄

src/renderer/components/archive/entry-view.js
ph1p: Oh cool, didn't know that :)

src/renderer/components/tree-view/tree-label-edit.js
ph1p: I'll change it ;) Don't know if it's better to write `propTypes` in- or outside class. What do you prefer?

perry-mitchell: I think `static propTypes` within classes is what we're going for now - last I spoke with @sallar :)

src/renderer/components/archive/entry-view.js
perry-mitchell: Yeah it's great! Much nicer imo to read just XML..

perry-mitchell:
Nice @ph1p! I'll still let @sallar check the code as this is quite a large update.

Thanks..

ph1p:
Ok :)

sallar:
Great job @ph1p

PR #401, opened 2017-11-15
Update Spanish locale file

multiservicio:


multiservicio:
Once is available I will revisit it as with the context it will be easier to spot texts to fine tune.

perry-mitchell:
Thanks @multiservicio !!

PR #430, opened 2017-11-18
Fix expand bug in tree view

ph1p:
After two redbull and one line of code, I've fixed a bug in the treeview. 

**Reproduce:**
nov-18-2017 19-02-08


perry-mitchell:
Nice work @ph1p 👌

ph1p:
*pew pew* your welcome. This line took me 2 hours :D

Issue #406, opened 2017-11-15
Locales: Nest some items


perry-mitchell:
Currently the locales are mostly hyphenated. Some ones, like **errors**, could be prefixed with such a namespace: `error.some-item`, for example.

Change some of the obvious locale property names to namespaced options.

ph1p:
I thought about nearly the same type of naming, this is why I added `-error`.

**Some proposals + examples:**
- archive - open-archive-file **to** archive.open-file
- entry - add-entry **to** entry.add
- group - new-group **to** group.new
- error - archive-not-found-error **to** error.archive-not-found
- info - archive-saved-loading-info **to** info.archive-saved-loading

perry-mitchell:
I think that you've got the idea, besides maybe _info_. In my opinion it'd be best to sort by logical area of the app. Errors are their own thing (all messages), but the other areas of the app are very easily distinguishable:

 * Archive(s)
 * Groups
 * Entries
 * Importing
 * Menus
   * Context menu on archives
   * Window menu
   * Context menu on entries etc.
 * Connections (webdav/dropbox etc.)
 * Menu actions (hide, reveal etc.)
 * User actions (Confirm, Cancel, Yes, No etc.)

hugo-cardenas:
Hi guys!
Finally I get to give a hand with Buttercup. I thought to pick this as a first simple and boring task, to get going with an easy one and find my way around the project.

perry-mitchell:
Hey @hugo-cardenas! Welcome. That would be awesome! Let's get through that current PR and you could most definitely take this item :)

perry-mitchell:
This is now ready @hugo-cardenas, if you get time. I guess there are already some nested errors, but I think there's a lot that could be categorised. This can obviously be a partial PR due to the large amount of items that may need changing.

hugo-cardenas:
Hi! I created already a PR (#446) in order to discuss better the format as I change things.
I nested the items for the window menu in the English config for now.

Are you guys ok with the structure like this?

Also, I noticed there were some items used in multiple contexts (e.g. `hide`, `copy`, etc). At the moment, for those used from menu + somewhere else, I kept them duplicate, in the new menu structure, and in their original place, used somewhere else.

I believe it would be better not to reuse items in different contexts, even though they might happen to contain the same value.
If later some UI element is removed/changed, when you remove/change the locale items with it, it's nicer if you don't risk breaking other things which use the same item. 
What do you think?

[...]

PR #434, opened 2017-11-20
Replace react-intl with i18next

### LINKED ISSUE #417: Detect system language
It would be nice if Buttercup could detect the user's OS language and set the language based on that automatically (if available) or fall back to English.

This package can be used: https://github.com/sindresorhus/os-locale

ph1p:
Hi! Me again.

I've decided to replace react-intl with i18next.
This Issue (#417) is solved by using  `app.getLocale()` in the main part.

**But why???!!!**
I was not satisfied with my solution and dreamed about it last night. 😊

**Some advantages of i18next:**
- plugin support
- better integration with react-native (buttercup mobile)
- better fallback solution
- easier to use with electron
- better configuration possibilities
- formatting (https://www.i18next.com/formatting.html)
- namespace support (#406)
- split translation files
- seperators (https://www.i18next.com/configuration-options.html#misc)

What do you think? (:

Fixes #417


perry-mitchell:
Hi @ph1p - Thanks again for the huge amount of work. We of course need to check this over and that might take a bit.

I'm not completely sold on i18next.. and for one the react integration looks non-existent. The `{t(..)}` insertion into JSX looks quite bland compared to the dedicated JSX components in react-intl. At least I think it would be advantageous for us to wrap it in a JSX component if we're to go with this solution - thoughts @sallar?

I'll have to have a proper review a bit later. In the future I would recommend starting an issue with a discussion before changing anything to this degree. If a feature takes us in a direction we're not 100% happy with it could end up being rejected that would of course be a massive waste of work. Just a friendly point to ask first if it's a large change like this 🙂

ghostd:
Just for the discussion: did you take a look at [js-lingui](https://github.com/lingui/js-lingui)?

perry-mitchell:
@ghostd I do like the way js-lingui works, especially with these items: `<Trans id="msg.simple">Hello {name}</Trans>`

I think it's important that the translations in react are:

 * JSX elements, not JS insertions
 * Easy to read and understand without having to jump to the locale file
 * Shows the default in some easy way

If we're to stick with i18next or a similar framework, I would recommend using something like the above. I think having the default as the `children` for the translation element is a very good idea.

ph1p:
Hey @perry-mitchell 
There is a react component. Don't know why I haven't used it. I'll replace it.
```
<Trans i18nKey="userMessagesUnread" count={count}>
  Hello <strong title={t('nameTitle')}>{{name}}</strong>, you have {{count}} unread message. <Link to="/msgs">Go to messages</Link>.
</Trans>
```
Source: https://github.com/i18next/react-i18next

> In the future I would recommend starting an issue with a discussion before changing anything to this degree.

I wanted this code as base, because I don't want to start a discussion, when I don't know the library.

> If a feature takes us in a direction we're not 100% happy with it could end up being rejected that would of course be a massive waste of work. Just a friendly point to ask first if it's a large change like this.

This wasn't a massive waste of work for me. I wanted to compare `react-intl` with `18next`. It's not bad if it's rejected by you. 
And jep next time I'll open an issue. Sorry :/

perry-mitchell:
@ph1p I would actually hold off on changes at the moment mate.. I think we need to have a bit more of a discussion on this, this week, to see what we want to do.. If you wouldn't mind waiting on a decision, we could choose something and just stick with that. We don't want to waste your valuable time and effort..

But hey! This is really exciting.. I'm glad that you've gotten us to where we can make these choices. Let's see how we go. Maybe we start a discussion on Spectrum..

locales/fr.json
```diff
@@ -1,95 +0,0 @@
-{
```
sallar: Is it possible for you to use git mv instead of mv for moving template files so the git history on the files wont be lost? It seems that fr.json is deleted and created again.

locales/fr/translation.json
```diff
@@ -0,0 +1,141 @@
+{
```
sallar: Also what is the difference between template.json and translation.json?

src/main/actions.js
```diff
@@ -50,8 +51,23 @@ export function setupActions(store) {
...
-    i18n.setup(store);
+  ipc.on('change-locale-main', (e, lang) => {
+    let locale = lang;
+    if (!locale) {
+      const win = getMainWindow();
+      // set system lang
+      locale = app.getLocale().split('-')[0] || 'en';
+
+      // check if its available
+      if (Object.keys(languages).indexOf(locale) >= 0) {
```
sallar: This block is wrong. It should be checking if its "not" available, then use default:

``` js
if (!(locale in languages)) {

File: src/renderer/components/empty-view.js [hunk 4]
```diff
@@ -44,46 +44,54 @@ const Title = styled.h3`
...
-          defaultMessage="Unlock an archive to begin ({os})."
-          values={{
-            os: `${isOSX() ? '⌘' : 'Ctrl'}+1`
-          }}
-        />
+        <Interpolate
+          i18nKey="unlock-archive"
+          os={`${isOSX() ? '⌘' : 'Ctrl'}+1`}
+        >
+          Unlock an archive to begin ({`${isOSX() ? '⌘' : 'Ctrl'}+1`}).
```
sallar: Cant you just do `{{os}}` here?

src/shared/i18n/index.js
```diff
@@ -1,109 +1,46 @@
...
+
+i18n.init({
+  fallbackLng: 'en',
+  resources,
+  react: {
+    wait: true
+  },
+  ns: ['common'],
+  defaultNS: 'common',
+  keySeparator: '>',
```
sallar: Key scanner and ns separator are different here than in the scanner config. I kind of prefer the ones in the scanner. : and .

test/app.js
```diff
@@ -62,13 +64,15 @@ test('test input focus', async t => {
   buttons = await app.client.elements('button');
   await app.client.elementIdClick(buttons.value[3].ELEMENT);
 
-  await sleep(1000);
+  await sleep(2000);
```
sallar: Are you increasing numbers just to be sure or the i18next performance is lower?

locales/fr/translation.json
ph1p: translation.json will be used in the app.
i18n-scanner extracts keys from i18n.t(), t(), <Trans> and <Interpolation> -> i18n-scanner searches in template.json for existing keys and translations -> translation.json is generated based on template.json and found keys will be appended. Non existing keys are automatically set to english translation. I can remove this or integrate it in a seperate folder.

test/app.js
ph1p: AppVeyor build failed. I'll check the performance.

src/shared/i18n/index.js
ph1p: Ok I'll replace them with the scanner separators.

src/main/actions.js
ph1p: Even better solution (: thanks

locales/fr.json
ph1p: Jep. I'll try my best.

ph1p: done

locales/fr/translation.json
perry-mitchell: Yeah this is a bit confusing. Does it mean that we currently have duplicates? I think one .json file per language should be sufficient right?

src/shared/i18n/index.js
```diff
@@ -3,23 +3,23 @@ import i18n from 'i18next';
...
     name: 'Français',
-    common: require('locales/fr.json')
+    common: require('locales/fr/translation.json')
   },
   ru: {
     name: 'Русский',
-    common: require('locales/ru.json')
+    common: require('locales/ru/translation.json')
   }
 };
```
perry-mitchell: This will need updating for Italian.

ph1p:
I'll continue the i18next feature today early in the evening and you're right @perry-mitchell. Currently these JSONs are nearly identical. I don't like it either. I'll think about a logical structure and push all my changes later.

perry-mitchell:
@ph1p @sallar I do like the examples that I saw on their homepage a while back: [here](https://www.i18next.com/getting-started.html#extended-sample). Having the namespaced files is nice imo (errors, menus etc.).. but ultimately not necessary. It seems that it's fine to have just 1 file for the translations as with react-intl..

ph1p:
Hey! Spectrum is currently under heavy load :/ I'm working hard to get the i18next feature finished.
I've written a <Translate> component and appended an i18next parser to the **i18next-scanner.config.js** file for it. I've done that, because we've two components <Trans> and <Interpolation> from react-i18next.

**<Trans>**
```
<Trans i18nKey="date" parent="span">
    Date <strong>{{date}}</strong><br /> etc.
</Trans>
```
You can't pass the value `{{date}}` to it and the interpolation brackets are detected as object.
The i18next scanner is creating something like this:
`Date <0>{{date}}</0> etc.` (jep,  <br /> was not recognized)
I think that's very confusing.

**<Interpolation>**
```
<Interpolate i18nKey="connect-to-wedav" title={title}>
    Connect to {{ title }} Server
</Interpolate>
```
You can pass values as props, but not use them, because of the same problem with objects above.
Additionally you can't automatically parse for <Interpolate> translations.

**<Translate>**
```
<Translate
  html
  i18nKey="i18n-key"
  values={{
    name: 'Peter'
  }}
  defaultText={`
    Hi {{name}}
 `}
/>
```
This is my component. All necessary attributes can be passed as props. This component is small an uses directly all methods from i18next.

What do you think?

---

Currently the **locales/** folder looks like this:

- locales/
  - en.json
  - de.json
...

With i18next:

- locales/
  - en/
    - translation.json (namespace)
    - error.json
  - de/
    - translation.json (namespace)
...

I've removed all template.jsons. **translation** is the namespace. If we write i18nkey="error:my-error" a new file will be created with that name and we have to add it to **shared/i18n/index.js**. But we can also indent the error messages and have only one file.

- error:entry-not-found (namespace, nsSeparator)
- error.entry-not-found (indent, keySeparator)

Should we indent or namespace translations?
Does the folder structure fit like this?

perry-mitchell:
Hi @ph1p, I've summarised my points below:

 * ~~I'm not sure that I love the `<Translate>` component.. Why not use [react-i18next](https://github.com/i18next/react-i18next)'s `<Trans>` component? It'd be nice if we didn't have to invent anything new for this.~~
 * ~~I think the default text should be a child of the component, not a prop. Passing to `defaultText` on the `Translate` component is an antipattern, as some defaults can be long sentences and should be passed as `<Translate>This is the default text</Translate>`, for example.~~
 * I think you've had some really good ideas on the `<Translate>` component, but I'm still not completely comfortable with how it might look. I'd like something solid that we can use throughout the entire Buttercup stack, something that reads well, but this is still a bit cumbersome. What about ditching the default altogether? We don't really need it as it **should** fall back to English.. and the default value only serves to complicate the component. Having something like `<Trans i18nKey="userMessagesUnread" count={count} />` would be amazing, in my opinion, without the default value (avoids all the complex react object stuff). What do you think? /cc @sallar 
 * I don't think we should do any namespacing in this PR.. it's already huge 😃 - Let's stick with the single file for now. Maybe something like `base.json` for a name would be nice.
 * I like the folder structure you proposed - It will suit us nicely when we do begin namespacing.
 * What do you mean by indenting? Such as with the phrase "But we can also indent the error messages and have only one file." ?

ph1p:
> What do you mean by indenting?

We can write error:entry-empty or error.entry-empty

First one will create a file called **error.json**:
```
{
    "entry-empty": "Text"
}
```

And the second one creates a file called **translation.json**:
```
{
    "welcome": "Welcome"
    ...
    "error": {
        "empty-entry": "Text"
    }
}

```

base.json is a good name :)

---

> I think you've had some really good ideas on the <Translate> component, but I'm still not completely comfortable with how it might look.

I will edit it and push it later.

---

> Why not use react-i18next's <Trans> component? 

This is a good question. 

```
render() {
  const name = 'Peter';
  const count = 5;

  return (
    <div>
      <Trans i18nKey="userMessagesUnread" count={count}>
        Hello <strong title={t('nameTitle')}>{{ name }}</strong>, you have
        {{ count }} unread message. <a href="/msgs">Go to messages</a>.
      </Trans>
    </div>
  );
};
```
We have to set all variables before.

Transforms to:

```
{
  "userMessagesUnread": "Hello <1><0>{{ name }}</0></1>, you have <3>{{ count }}</3> unread message. <5>Go to messages</5>."
}
```
I don't know if i18next filters this in a non react context.

**Some Problems I have had:**

This will cause in an error after auto generate translations. I think this is because of the self-closing tags.
```
<Trans i18nKey="userMessagesUnread" count={count}>
  Hello <strong title={t('nameTitle')}>{{ name }}</strong>, you have<br />
  <br />
  {{ count }} unread message. <a href="/msgs">Go to messages</a>.
</Trans>
```

react-i18next does not detect the full translation.
```
<Trans i18nKey="userMessagesUnread" count={count}>
  Hello <strong title={t('nameTitle')}>{{ name }}</strong>, you have<br />
  <br />
  {{ count }} unread message. <a href="/msgs">Go to messages</a>.
</Trans>
```

Output:
```
<div>Hello <strong title="nameTitle">Peter</strong>, you have<br> unread message. unread message. ..</div>
```

Transforms to:
```
{
  "userMessagesUnread": "Hello <1><0>{{ name }}</0></1>, you have<3></3> <5>{{ count }}</5> unread message. <7>Go to messages</7>.",
}

```

Or prettier adds `{' '}` and the translation does not fully render:
```
<Trans i18nKey="userMessagesUnread" count={count}>
Hello <strong title={t('nameTitle')}>{{ name }}</strong>, you have{' '}
{{ count }} unread message. <a href="/msgs">Go to messages</a>.
</Trans>
```
Output: Nearly the same as above.

Frequently translations
---

> I think the default text should be a child of the component

Yep.

---

Additionally I will take a look at jslingui and see how all of this stuff is done there 🤗

---

As I said I'll push it later and we can review all solutions. We're not in stress, because a temporary solution is already there and these decisions are also important to other projects like react-native, core or website. But I think you both know that already (:

perry-mitchell:
@ph1p Just on my last notes - Just wanted to reiterate that in my opinion, it might be simpler having the `<Translate>` component with **no children**, like you suggested. Also, it might be easiest there if we have **no default** specified in code. Relying on the English fallback should be enough for all cases.

sallar:
@ph1p Hey man, sorry for the delay, we have been busy a bit with other things, but finally back on this. I'll review this tonight :)

ph1p:
@sallar thanks :) I was also busy the last few days.

config/i18next-scanner.config.js
```diff
@@ -0,0 +1,62 @@
...
+  console.log('default langauge not found');
+}
+
+module.exports = {
+  options: {
+    debug: true,
+    func: {
+      list: ['i18next.t', 'i18n.t', 't']
+    },
+    lngs: ['en', 'de', 'fr', 'ru', 'es', 'it', 'fa'],
```
perry-mitchell: So there's languages specified (hardcoded) also in `src/shared/i18n/index.js`.. Would it be easiest to store them both somewhere together and use them in both places?

perry-mitchell:
Sorry @ph1p still a few points I'd love to get cleared up before merging. Thanks!

ph1p:
Ok I'll do my best

ph1p:
I set all missing translations as an empty string. All translators see what's missing now.
I've added also a config file with all translations, a type array and the fallback language.

perry-mitchell:
> missing translations as an empty string

Ah, I meant removing them entirely. My thinking was that if there are more texts added in the future that the default will be that some items will not be set in all language files. We can of course remove these too, you don't need to do it now. Will take a look myself tomorrow.

ph1p:
> Ah, I meant removing them entirely.
Sorry I forgot to mention, that this was an suggestion of mine ^^.  I thought this were better, because when I take a look at all translations and everything is filled out, I don't start a PR.
All empty translations are using the fallback language.

perry-mitchell:
> when I take a look at all translations and everything is filled out

Yes but when someone adds new functionality, I don't think it's fair to ask them to add the entries to all language files. Like react-intl, the items should be added to the base/default language and not to the others. The missing translations should be provided by running a CLI application or the like..

I think removing them is best as this will most likely be how we move forward with new features. Having empty entries I feel is still a bit misleading..

ph1p:
If you add new functionality, you can run:`npm run i18next-parse-lang`. The scanner will find all keys and append them to the right file. All existing keys are not touched by this.
---
But ok I'll remove them ;)

perry-mitchell:
Well I'm not so familiar with i18next. If it creates blank entries automatically, then we probably want to do whatever the default is. Something that devs don't need to think about if they're adding an unrelated feature. That's kinda what I'm getting at 😅

perry-mitchell:
But let's go with what you have now.. this is fine. Thanks again!

ph1p:
Removed them already ^^. Try it by yourself. Add a fake locale, run the command above and you'll see the result in locales. No one has to think about it. Devs add the translation methods to their features, scanner finds them and the translator fills out the empty string.
I set this flag `returnEmptyString: false` in `/shared/i18n/index.js` to handle this behavior.

sallar:
@perry-mitchell @ph1p Thanks guys. merging! 🎉

PR #446, opened 2017-11-30
Nest locales config

### LINKED ISSUE #406: Locales: Nest some items

hugo-cardenas:
Fixes #406


src/main/menu.js
```diff
@@ -14,23 +14,25 @@ import { getMainWindow } from './utils/window';
 import i18n, { languages } from '../shared/i18n';
 import pkg from '../../package.json';
 
+const label = (key, options) => i18n.t('menu.window.' + key, options);
```
perry-mitchell: Nice 👍

perry-mitchell: This could be maybe a bit neater with a template string however: 

```javascript
i18n.t(`menu.window.${key}`, options)
```

locales/en/base.json
```diff
@@ -1,33 +1,84 @@
 {
+  "menu": {
+    "window": {
```
perry-mitchell: Not 100% sold on the 3 level deep stuff here.. Wondering if we could make it simpler by just having something like `"appmenu"` or something like that? I don't hate it I'm just not sure if we need the extra level, especially seeing as there is another `"window"` group further into the nested items.

locales/en/base.json
```diff
@@ -1,33 +1,84 @@
 {
```
perry-mitchell: Of course we also need to transform the other languages when this is decided 😊

perry-mitchell:
Made some comments @hugo-cardenas, in the way of a review. Easier to chat about the items that way 😁

locales/en/base.json
hugo-cardenas: Ofc

locales/en/base.json
hugo-cardenas: Fair enough. `app-menu`, `archive-menu`, as top level items are still quite easy to find.

perry-mitchell:
@hugo-cardenas Looking good!

hugo-cardenas:
I'm done now with the app-menu category for all languages. Should we wrap-up this PR for a start, in order to keep them smaller? I can continue with the others after that.

Also:

- Taking the english version as the master config, I did JSON diffs with each of the others in order to double check any possible mistakes, and noticed there are several entries missing from most of the other languages. I synced them, meaning that I added all the missing entries with the english value to each of the other languages, but now different people should review those again and translate them.
- Whenever new english entries are added, I think the best would be to initially always add them to all the other languages even if just in english, so later it's easier for translators to just look at their corresponding language file and change the english values to their language?

perry-mitchell:
> I synced them, meaning that I added all the missing entries with the english value to each of the other languages, but now different people should review those again and translate them.

> I think the best would be to initially always add them to all the other languages even if just in english

We shouldn't set items to English by default.. The intention is that they should always be unset if not translated. Setting them to English makes it unclear as to what should be translated. There are also CLI tools that help identify what items need translating, and these would effectively not work if we set all items to English by default. The translation system will **automatically default to English** if an item is not found in the current locale.

You're not the first person to suggest this, so maybe we have to add this to our docs 😊 - Apologies for the confusion there.

perry-mitchell:
I added some docs about this for future contributions: #448. react-intl (our previous system) and i18next (our current system) both support the same behaviour: Try to fetch the localisation from the current locale, and if it's not there, get it from the base locale (en.json).

hugo-cardenas:
@perry-mitchell Oh, my bad. I saw that the locale keys were shown when not found in the config, but did not notice that before that it will try to fall back to english.

I removed now the english values from all the other files.

> There are also CLI tools that help identify what items need translating

Would it be even a good idea to install something like `json-diff` to devDependencies and add the required specific command to the `Translations` instructions in `CONTRIBUTING.md`? 
```
./node_modules/.../json-diff ./locales/en/base.json ./locales/my-lang/base.json
```
Like this the effort is minimal for those willing to help with the different languages.

perry-mitchell:
I agree on adding the command. We can add it to the package.json.. I’ll try to find it from another PR I saw it in. It’s a specific tool for use with the i18n package we use.

hugo-cardenas:
Ok @perry-mitchell, after all, I can just go forward with the whole task in this PR.

I nested the whole english config and updated all the code usages. 
If you have any comments on the JSON structure, it would be best to make all the changes now first, and when that's ready, I'll update all the other locale files.

perry-mitchell:
@hugo-cardenas I really like the nesting you've done.. It's good from my perspective if you continue like that. @sallar ?

hugo-cardenas:
Ok, now everything should be **finally** ready!

In the end, I was so annoyed by copying the format from english to others, that only with 3 languages left, I wrote a small script for copying from flat to nested format. Should have done that in the very beginning, and just fully reformat manually the whole english file first of all 😅

perry-mitchell:
Good work @hugo-cardenas! Thanks so much for pushing through on this.. it was quite a fiddly task. I'll defer to @sallar to merge once he gives it a quick check (he's definitely more familiar with this project).

hugo-cardenas:
🎉 Thanks @sallar!

PR #450, opened 2017-12-04
Add https prefix to incomplete URL when opening in browser

### LINKED ISSUE #440: Opening URL in browser does not work on incomplete URLs
The "Open URL in Browser" content menu option does not work when the URL is one of the following, for example:

 * `google.com`
 * `www.google.com`

But will for the following:

 * `https://google.com`
 * `http://google.com`

All URLs, no matter what the value, should be launchable..

hugo-cardenas:
Fixes #440

### DIFF
--- a/src/renderer/system/utils.js
+++ b/src/renderer/system/utils.js
@@ -12,6 +12,9 @@ export function readClipboard() {
 }
 
 export function openUrl(url) {
+  if (!/^http(s)?:\/\//i.test(url)) {
+    url = `https://${url}`;
+  }
   shell.openExternal(url);
 }
 


hugo-cardenas:
@perry-mitchell Would you consider at some point adding some testing utility which would allow to stub/mock modules?

I tried to add a tiny simple test case to cover this, but the function modified is tied to `electron.shell`, so it's not possible without mocking it.

In this case the test is very dumb and not a big deal, but I think being able to test some other bigger pieces of code in isolation from other modules would be useful.

I've used module mocking functionality from Jest, which works great, but I don't have experience with other standalone modules providing this. I think for example [rewire](https://github.com/jhnns/rewire) and [proxyquire](https://github.com/thlorenz/proxyquire) are popular ones.

perry-mitchell:
Absolutely @hugo-cardenas - But I would suggest that we take the cleanest route here. I've used rewire before and to be honest it's a bit of a pain to use. It also uses a bit of its own magic to overwrite values in-script which doesn't work for every declaration type - for these reasons and more I'd prefer not to use it.

Ideally we'd pass dependencies into functions which *really* need testing, but this is untidy and obviously doesn't test all of the branches. Whatever we do in the end will still be a sacrifice but as long as we cover the most important areas..

hugo-cardenas:
Thanks @sallar!

@perry-mitchell I definitely agree that "rewiring" the modules for tests is a bit of a dirty way of mocking things, but might be useful sometimes when it's really hard or costly to refactor bottom-up some existing code in order to pass in dependencies.

In another order of things, I've found myself liking a lot this one pattern, using the `index.js` default import functionality to hide away dependency injection.

E.g. I mean if I would create some module `foo`, my folder structure would be
```
/src/foo/foo.js
/src/foo/index.js
```
Where `foo.js` is the real module, which accepts dependencies as arguments.
And `index.js` is just the wiring, fetching those dependencies and passing them to the module, returning the module itself.
Then from somewhere else I just `import 'src/foo'`, so the whole dependency wiring is hidden.

perry-mitchell:
@hugo-cardenas Having an extra file for the wiring is **a huge amount** of overhead though.. The mobile app has 10s of modules and the browser extension possibly over a hundred by now.. If we adopted such a pattern the file overhead would be bonkers.

I agree that we can give a little in terms of coding style to make testing easier, but that approach sounds a bit drastic imo.

Webpack has a loader called [inject-loader](https://github.com/plasticine/inject-loader) which allows for overwriting _imported_ items. This helps in most cases where modules don't rely too much on internal functions (it can't rewrite functions within the same module file). For those, it's better to look at the test in the following way imo: You're testing the module, and thereby testing what methods it exposes.. if a method cannot be mocked within the module, test the items that use it and treat it as embedded functionality (as if it were in the function you're currently testing).

Using inject-loader might work well for this as we already use webpack. It won't help with the node side of things though.

perry-mitchell:
More on this, due to the fact that I'm encountering the same thing in browser extension right now, core-dependencies (browser/electron etc.) should be moved into a separate module so using something like inject-loader can easily help mock those core methods.

hugo-cardenas:
I agree it's a lot of extra files, basically doubling in all modules with deps, if you would take that pattern as standard.

Personally I don't see this as a big problem - if I would have for example a folder `tools/` with 5 different module files inside, there would be instead 5 module folders inside `tools/`, each with `module-foo.js` + `index.js`. But I don't need to think about this when using them, only when implementing. After that, they are just hidden inside the module folders. And `index.js` files are tiny. Then suddenly you can just easily test every single module you write.

But of course, this is your (both) project and it's your call. I did not intend either to suddenly set a whole bunch of different practices. Just my 50ct, maybe you could find it suitable at some point 😊  

I'll take a look into `inject-loader`, thanks. I understand the main difference with those `rewire` and others is that it's "less tricky"? So it just applies some extra rules at compilation time instead of some other "dirtier" magic to replace the modules.

PR #454, opened 2017-12-10
Entry icons

### LINKED ISSUE #16: Entry icons
Support for custom entry icons

hugo-cardenas:
Fixes #16


package.json
```diff
@@ -8,7 +8,9 @@
     "clean": "rimraf ./app/node_modules",
     "format": "prettier --write '{src,config,test,locales}/**/*.{js,json}'",
     "precommit": "lint-staged",
-    "test": "cross-env NODE_ENV=test ava && npm run lint",
+    "test": "npm run test:unit && npm run test:gui && npm run lint",
+    "test:gui": "npm run build && cross-env NODE_ENV=test ava test/gui/**",
```
hugo-cardenas: This is slow... but the GUI test really runs against the built app, so without running `build` it isn't really testing the current code (plus the 1st time it did not work until you manually run build yourself)

package.json
```diff
@@ -134,6 +130,7 @@
   },
   "devDependencies": {
     "@buttercup/channel-queue": "~0.2.1",
+    "@buttercup/iconographer": "git://github.com/hugo-cardenas/iconographer.git#fc099568b7b5432d9a336f8c774e057fdce66d11",
```
hugo-cardenas: To be updated when Iconographer PR is merged

src/main/lib/icon/storage/IconFileStorage.js
```diff
@@ -0,0 +1,103 @@
...
+
+export default class IconFileStorage extends StorageInterface {
+  constructor(path) {
+    super();
+    this.path = path;
+  }
+
+  _buildKeyPath(iconKey) {
+    // TODO No extension?
+    return path.join(this.path, sanitize(iconKey)) + '.ico';
```
hugo-cardenas: TBD

src/main/menu.js
```diff
@@ -13,6 +13,9 @@ import { checkForUpdates } from './lib/updater';
 import { getMainWindow } from './utils/window';
 import i18n, { languages } from '../shared/i18n';
 import pkg from '../../package.json';
+import electronContextMenu from 'electron-context-menu';
+
+electronContextMenu();
```
hugo-cardenas: This is something I'm not sure if you will agree with. But I found for myself having `Inspect` option in development as essential when tweaking the UI. Using [this module](https://github.com/sindresorhus/electron-context-menu) was the easiest way I now of getting that menu there.

It adds also `Cut`, `Copy` and `Paste` for text, whenever it would suit and there isn't a Buttercup menu overriding it. I totally understand if you don't want to have these, please tell me what you think.

src/renderer/styles/img/lock-black.svg
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="iso-8859-1"?>
```
hugo-cardenas: You might want to use another icon of yourselves as the default, maybe even a Buttercup icon?
Just in case this would stay, it comes from http://flaticon.com, so would need to give credit on the `README`

src/shared/buttercup/entries.js
```diff
@@ -50,77 +51,132 @@ export function validateEntry(entry) {
...
+  if (icon) {
+    entryObj.icon = icon;
+  }
+
+  return entryObj;
+}
+
+async function fetchIcon(entry) {
+  // We move on whether this succeeds or not
+  // TODO Should maybe log it somewhere (but not alert the user - not an error)
```
hugo-cardenas: Iconographer error flow maybe could be improved (we probably don't want to show Iconographer errors to the user, but it would be good to log them somewhere they can be at some point gathered)

src/shared/buttercup/entries.js
```diff
@@ -50,77 +51,132 @@ export function validateEntry(entry) {
...
+    await iconographer.processIconForEntry(entry);
+  } catch (err) {}
+}
+
+async function getIconForEntry(entry) {
+  // TODO Decide better differentiate between errors? (Icon missing vs iconographer failed)
+  try {
+    const buf = await iconographer.getIconForEntry(entry);
+    if (buf) {
+      return `data:image/png;base64,${buf.toString('base64')}`;
```
hugo-cardenas: Temporary hack to be fixed

perry-mitchell:
Thanks @hugo-cardenas - I'll review in a bit. On a side node, iconographer is out with your changes - version 0.2.0.

src/main/lib/icon/storage/index.js
```diff
@@ -0,0 +1,9 @@
+import IconFileStorage from './IconFileStorage';
+import electron from 'electron';
+import path from 'path';
+
+const app = electron.app || electron.remote.app;
+const iconPath = path.join(app.getPath('userData'), 'icons');
+
+const storage = new IconFileStorage(iconPath);
+export default storage;
```
hugo-cardenas: Btw, here a practical example of that `index` pattern we discussed.
If you still don't like it like this, I'm open to discuss other approaches.
IMO it's just a great way of hiding the wirings inside a folder + testing very easily the `IconFileStorage`

src/main/lib/icon/storage/IconFileStorage.js
perry-mitchell: I would recommend an extension of `.dat`, for instance, to prevent any arbitrary assignment of type (eg ico/png/jpg etc.).

src/main/lib/icon/storage/index.js
perry-mitchell: I'm still not sold on it.. I don't see immediate benefit from using it and I feel it overcomplicates the file structure of the application. In your example you return some-what of a singleton as you're instantiating the class and providing an arg to it.. but this could all be done in-module imo. For instance:

```javascript
let __sharedInstance;

class MyClass {}

export default MyClass;

export function getSharedInstance() {
  if (!__sharedInstance) {
    __sharedInstance = new MyClass(stuff);
  }
  return __sharedInstance;
}
```

The class itself is easily testable and mockable at this point. The only reason that you'd want to intercept `__sharedInstance` here is for integration testing. For that I'd argue that when integration testing, you should mock as little as possible to test the actual platform.. especially if you're going to the trouble of integration testing. Also, if you are going to have to change it to point elsewhere for some reason, you could inject said dependency upstream or provide a setter for the singleton. This latter way is obviously more ugly but integration testing sometimes needs a more complex harness for it to work.

Tldr: I don't think that, for the sake of integration testing, that it's necessary to add index files for directories 😅

Also, I may have completely misunderstood your point.. if so I apologise 😬

src/main/menu.js
perry-mitchell: I'll let @sallar check this one out. The Cut/Copy/Paste sounds good to me..

src/renderer/styles/entry-icon.scss
```diff
@@ -0,0 +1,17 @@
+@import './variables';
```
perry-mitchell: I think this can be just `@import 'variables';`

src/renderer/styles/img/lock-black.svg
perry-mitchell: Thanks @hugo-cardenas - Good idea. I think we'll definitely use our own. What icon do you think we should use @sallar?

src/shared/buttercup/entries.js
perry-mitchell: I would recommend using the [batched approach](https://github.com/buttercup/iconographer#managed-batch-icon-fetching) for the desktop app.. Immediate when seeing an icon for the first time, call `processIconForEntry ` and then for every entry just `getIconForEntry`. All `null`s should just default to the default icon.

perry-mitchell: Basically, the default icon should always be visible and only replaced by a successfully fetched icon.

test/unit/main/lib/icon/storage/IconFileStorage.test.js
```diff
@@ -0,0 +1,127 @@
+import test from 'ava';
+import path from 'path';
+import pify from 'pify';
+import fsLib from 'fs';
+import tmpLib from 'tmp';
+import IconFileStorage from '../../../../../../src/main/lib/icon/storage/IconFileStorage';
```
perry-mitchell: Lol.. @sallar I wonder if there'd be any benefit from bundling into a single file for the app.. maybe before testing. Not sure what we currently do here.. but we do bundle at least the front-end right?

src/main/lib/icon/storage/IconFileStorage.js
sallar: ^

src/main/lib/icon/storage/index.js
sallar: Im good with both approaches

src/main/menu.js
sallar: @hugo-cardenas @perry-mitchell I like it too. If it doesn't interfere with other menus that we have

src/renderer/components/archive/entry-icon.js
```diff
@@ -0,0 +1,20 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import styles from '../../styles/entry-icon';
+import defaultIcon from '../../styles/img/lock-black.svg';
+
+const EntryIcon = props => {
```
sallar: Instead of doing this, we can perhaps do styled-components:

```js
const EntryIcon = styled.div`
  background-image: `url(${props => props.icon ? props.icon : defaultIcon})`;
`
```
etc. Since we are trying to use styled-components in the project instead of css

src/renderer/styles/img/lock-black.svg
sallar: @hugo-cardenas can I get a screenshot of this in action?

src/shared/buttercup/entries.js
```diff
@@ -50,77 +51,132 @@ export function validateEntry(entry) {
...
-    : Promise.reject(errorMessages.join('\n'));
+  if (errorMessages.length > 0) {
+    throw new Error(errorMessages.join('\n'));
+  }
+
+  return filterEmptyEntryValues(entry);
 }
 
-export function loadEntries(archiveId, groupId) {
+export async function loadEntries(archiveId, groupId) {
```
sallar: how fast is this function going to be? Can't we load the icons async?

src/shared/buttercup/entries.js
sallar: @hugo-cardenas we have a module called `electron-log` which we are using to log stuff to user's userData directory

test/unit/main/lib/icon/storage/IconFileStorage.test.js
sallar: lol I have no idea what is going on here! :D @perry-mitchell @hugo-cardenas

src/main/lib/icon/storage/IconFileStorage.js
perry-mitchell: Maybe even `.icon.dat`, for instance..

src/main/menu.js
hugo-cardenas: It does not interfere, actually it is there unless your own menus overwrite it. This is due that I just added it as a base here, and whenever you would have a menu it just replaces it. If you are ok with this, good, it was just the least effort alternative.

If you would even like to merge it with your own menu, I could also do that, probably.

src/renderer/styles/img/lock-black.svg
hugo-cardenas: Maybe you have already seen it but:
screen shot 2017-12-11 at 22 23 26

src/main/lib/icon/storage/index.js
hugo-cardenas: @perry-mitchell I see your point in that actually thanks to JS modules supporting default + secondary exports, it is possible to use this in different ways to define in one single JS file both the "pure" module (unit testable) + the instantiated one (dependent on specific config or on other modules which you would want to replace in your test).

I guess in the end it comes to a matter of taste of how to apply the same idea. I like better somehow to separate very clearly in different files the pure module from it's specific instantiation/configuration.

In this case anyway, the point of this was to expose the pure non-configured `IconFileStorage` in one place (so in the test I pass my own specific path) and in another place provide both its real config (the Electron path) and make it a singleton.
It is totally about unit testing it (considering it unit even if using the filesystem).

Because here it is really not going to be used anywhere out of Iconographer, I could actually remove the `index` and do that part inside the `iconographer.js` itself. What about that?
But otherwise, an alternative for this could be what you suggest.

hugo-cardenas: Pd. I actually love how in JS you have such freedom of going with so many different approaches in this aspect 👍

hugo-cardenas: Pd2. I love discussing these meta aspects of code... maybe you noticed 😀

test/unit/main/lib/icon/storage/IconFileStorage.test.js
hugo-cardenas: Well, I just went with replicating source directory structure for tests... which unfortunately leads to this.
Alternatives to this ugliness:
- Define a different test directory structure
- Place tests together with code
- Use some utility or config to allow absolute imports from root, so that this would be something like the following, no matter where you call it from:
```javascript
import foo from 'src/main/lib/icon/storage/IconFileStorage'
```

@perry-mitchell Not sure I understand what you mean in your comment.

src/renderer/styles/img/lock-black.svg
perry-mitchell: Wow! I love it.

I would perhaps reduce the gap between the icons and the text a _tiny_ bit.. and maybe center the icon a bit more.. but it's looking very elegant.

src/shared/buttercup/entries.js
hugo-cardenas: @perry-mitchell I actually did not really get the point of having Iconographer both return a Promise and emitting Events. Probably I'm missing something.

Now I went down the rabbit hole and I still don't see it, maybe you can clarify this for me.

- So `Iconographer`.[processIconForURL](https://github.com/buttercup/iconographer/blob/master/source/Iconographer.js#L139) first enqueues a task fetching the wanted icon, and when this task is completed, it enqueues another task which stores the icon data using the storage.

- I see that `Channel`.[enqueue](https://github.com/buttercup/channel-queue/blob/master/source/Channel.js#L122) returns the task promise, and this `Task`.[queuedPromise](https://github.com/buttercup/channel-queue/blob/master/source/Task.js#L45) actually resolves when the task is [completed](https://github.com/buttercup/channel-queue/blob/master/source/Task.js#L98).

- So going back up the rabbit hole, the Promise returned by `Iconographer.processIconForURL` actually resolves when the icon has been downloaded and stored into disk.

From what I understand now, the event emission would be useful if I would want to react in some universal way anytime **any** icon gets fetched and stored by the Iconographer. This could be defined totally separate from the specific Iconographer usage.

But this is not the case, as whenever I would react to the icon being stored, it would mean I want to fetch it and update a very specific entry. This means that actually this reaction is totally tied to the Iconographer usage and to the context of a specific entry.

So in the end, using the EventEmitter instead of Promises basically replaces this code:
```javascript 
// Creating or updating entry here
// Dispatch already the basic create/update with the entry data
// Then: 
let icon = await iconographer.getIconForEntry(entry);
if (!icon) {
  await iconographer.processIconForEntry(entry);
  icon = await iconographer.getIconForEntry(entry);
}
// Set icon to entry and dispatch entry update
```

with this code:

```javascript 
// Creating or updating entry here
iconographer.once('iconAdded', async url => {
  if (url === entry.url) {
    const icon = await iconographer.getIconForURL(targetURL);
    // Set icon to entry and dispatch update
  }
});
iconographer.processIconForEntry(entry);
// We can dispatch the basic create/update with the entry data here, 
// as we are not waiting for the icon update
```
...that was a long write. I hope you understand my doubts 😅

src/shared/buttercup/entries.js
hugo-cardenas: Yes yes. Missed this one. It should probably behave in the same way as the `createEntry` and `updateEntry`. First load the entry data fast without waiting for the icons, then fetch the icons for each entry (either disk or downloaded) and dispatch updates

src/main/lib/icon/storage/index.js
perry-mitchell: > Because here it is really not going to be used anywhere out of Iconographer, I could actually remove the index and do that part inside the iconographer.js itself. What about that?

> But otherwise, an alternative for this could be what you suggest.

Isn't that the same as what I suggested? 😅

I was suggesting having the singleton/instance getter inside the relevant module file.

> I actually love how in JS you have such freedom of going with so many different approaches in this aspect

Me too! It's sometimes a burden as it means there's more ways in which you can hurt yourself when it comes to maintainability, but usually it's a good thing.

> I love discussing these meta aspects of code... maybe you noticed

I did! Haha.. and so do we. Always great to have fresh eyes on what we're building too!

test/unit/main/lib/icon/storage/IconFileStorage.test.js
perry-mitchell: No.. was just laughing inwardly at how deep the structure is already. I don't think there are any action points here at this stage (at least not in this PR), but maybe something we can look at in the future..

src/renderer/styles/img/lock-black.svg
sallar: @hugo-cardenas you can use this lock from Google Material Icon Set which we also use in Buttercuo Mobile:
[ic-lock-48px.svg.zip](https://github.com/buttercup/buttercup-desktop/files/1556237/ic-lock-48px.svg.zip)

src/shared/buttercup/entries.js
hugo-cardenas: Or, should it actually wait **only** for those icons which are already stored in disk, and defer only the real slow part, downloading + storing + reading from disk icons for those entries which did not have a stored icon? (This 2nd part I actually missed to implement for `loadEntries`)

Not sure if there would really be a big noticeable difference for those entries with icons stored in disk (the usual case). If waiting, it would be a bit slower, but it would avoid tons of dispatch(updateEntry) for each entry in the loaded list.

src/shared/buttercup/entries.js
perry-mitchell: Yes I understand your point.. I think it's more clear the way you had it actually.  The event emitter I feel is still strong for some use cases, and I may use it in the mobile, but your use of the promise API is succinct. Let's stick with that!

hugo-cardenas:
Ok, finally I got to do all the fixes.
The threads are getting long, so I'll compile here the remaining important changes to be checked:
- Agreeing with [this](https://github.com/buttercup/buttercup-desktop/pull/454#discussion_r156710422)
- Still the original encoding concerns I wrote [here](https://github.com/buttercup/buttercup-desktop/issues/16#issuecomment-349604204). Why encode/decode? Atm it's not happening.
And what should be the image format [here](https://github.com/buttercup/buttercup-desktop/pull/454/files#diff-7f7c55a9e346e21ecb3728b18adc3b33R181)?
- I had exactly [this issue](https://github.com/i18next/react-i18next/issues/220) when a translation key was not found, that's why I did [this](https://github.com/buttercup/buttercup-desktop/pull/454/commits/a53e6c5c31dd2f6ae458e4effa1a6e125774cf2a) - the i18next backend is not even used at the moment.
- Agreeing with the appearance for now. I reduced the distance between icon and entry, but not sure what you meant @perry-mitchell by `center the icon a bit more`.
Also I played with having square icons, which maybe suits better many of them. Tell me which one you prefer. Atm they are circular.
Here is the current appearance (including also your updated lock icon):
  screen shot 2017-12-15 at 13 21 34
  Here with square icons:
  screen shot 2017-12-15 at 13 23 24

I think that should be it... almost there! 😅

sallar:
Thanks @hugo-cardenas. I kind of like the square more... Since most "favicon"s are actually square. In square one the logos are more visible and clear in my opinion. What do you think @perry-mitchell

sallar:
@hugo-cardenas the last point for me is to just maybe change the icons to square, otherwise im completely happy with this! I'll merge as soon as that's changed.

hugo-cardenas:
Thanks @sallar! I now pushed the change with the square icons.
But I think the 2nd point I listed should still be clarified.
* Why the encode/decode in `Iconographer`'s `StorageInterface`? Atm it's just not implemented in `IconFileStorage`. (Maybe @perry-mitchell knows more specifically the reason here)
* And what should really be the correct data URL, after getting the icon as `Buffer` from `Iconographer`? I don't really know what is the image type.
  This was just a temporary hack:

   ```javascript
   return `data:image/png;base64,${buffer.toString('base64')}`;
   ```

perry-mitchell:
> encode/decode in Iconographer's StorageInterface

The encoding overrides (optional) configure how the content is processed before reading and writing the icons in storage. It defaults to using buffers, which is fine for files in this case.

> This was just a temporary hack

I'd say this is pretty neat. I might ask that you put the base64 in a constant first, before inserting it into the string, but other than that it looks fine to me. Might be that libraries in npm that serve this purpose even do the same :)

sallar:
@perry-mitchell I think what @hugo-cardenas means is what is the original image type in the first place? is it definitly always PNG?

perry-mitchell:
@sallar No the image type can be anything.. For instance:

image

But it doesn't matter. Binary data is binary data, and base64 is base64.. Browsers (Chrome) will understand it when rendered as a CSS data URI, providing the icon is in a format that the browser would use anyway (guaranteed seeing as we fetched it from their webpage)..

sallar:
@perry-mitchell yes, so can't iconographer also return that part (the mime type) as well?

perry-mitchell:
@sallar Mime type is not needed here, and would be extraneous effort for nothing. The headers in the data tell the browser what format it is.. we're literally storing a _blob_ of data.

sallar:
ok cool then I think it's fine as it is 👍

perry-mitchell:
Ok.. after all that misunderstanding (sorry guys), I think i have a nicer way. Perhaps we merge this and I could make a new PR though?

sallar:
@perry-mitchell @hugo-cardenas Thanks guys! Im merging this! Hurray!

hugo-cardenas:
Noo damn. I'm so sorry @perry-mitchell @sallar I just actually broke the icons somehow in the last merging from updated master 😠  I will fix it and submit a new PR

sallar:
@hugo-cardenas sorry saw your comment too late. it's fixed now in master

hugo-cardenas:
Aah that's great, thanks @sallar!
I just noticed the issue here, I was missing those semicolons from the beginning but the new `styled-components` version is now strict on those... it was a time bomb 😆 

Btw, that means that the current `Prettier` version is not fixing `styled-components` syntax, so that should probably be updated, as the feature is available now in [1.8.0](https://github.com/prettier/prettier/releases/tag/1.8.0).

PR #457, opened 2017-12-11
Update French

vic1997:
I translated better in French (I'm French) and I added what was missing. There are still some things missing.


perry-mitchell:
This could be released with a larger update perhaps.

PR #461, opened 2017-12-17
Password field improvements

### LINKED ISSUE #456: Initially hide password when editing entry
Hi,

Love the app! I have a suggestion though: when clicking to edit an entry, the password is shown unobfuscated; I would prefer it to be hidden (just as a password field is) and having to click a button or icon to reveal, maybe even having to enter the master password again (as chrome does). I think it would be more secure this way, we don't always want to edit or see the password, just the title or some other field and having the password visible is insecure.

sallar:
Fixes #456 

password-input


perry-mitchell:
Looks good!

PR #465, opened 2017-12-18
Update icon -> base64 data URI processing

perry-mitchell:
Use a FileReader instance to create the URI


hugo-cardenas:
Neat! 👍

perry-mitchell:
Ok so I updated this.. turns out that FileReader with blobs **is** the right way to go, but we should write the result straight to the file cache. This data can be then used as is, because it contains the correct data URI when being processed by the `FileReader`.

src/main/lib/icon/IconFileStorage.js
```diff
@@ -19,13 +20,14 @@ export default class IconFileStorage extends StorageInterface {
...
-     */
-  decodeIconFromStorage(data) {
-    return Promise.resolve(data);
+  async encodeIconForStorage(iconData) {
+    const reader = new window.FileReader();
+    reader.readAsDataURL(iconData);
+    return new Promise(resolve => {
+      reader.onloadend = () => {
+        resolve(reader.result);
+      };
```
hugo-cardenas: Shouldn't you use rather `reader.onload` and `reader.onerror`, resolving and rejecting the Promise in each case?
Otherwise, it would just still store some bad result in case of failure (not even sure what would be the value), and would not retry anymore - when maybe retrying later to fetch the icon correctly might work and store the correct one.

If it rejects, it would cause Iconographer to reject, but that would not be a problem as we catch in `entries` and log if necessary.

src/main/lib/icon/IconFileStorage.js
```diff
@@ -37,25 +39,6 @@ export default class IconFileStorage extends StorageInterface {
...
-  }
-
-  /**
-     * Fetch all icon keys
-     * @returns {Promise.<Array.<String>>} A promise that resolves with an array of icon keys
-     */
-  async getIconKeys() {
-    const path = this.path;
-    return this._retryWithDirCreation(() => fs.readdir(path));
-  }
```
hugo-cardenas: I get the point that this is not used now in `Iconographer`, but maybe this means that this method should not be in the interface either?

perry-mitchell:
Yeah we were thinking that😊

hugo-cardenas:
Great, sorry if I commented on too quickly 😬

perry-mitchell:
@sallar @hugo-cardenas Ok it seems that iconographer wasn't really 100% for show time, so I'm going to take it out of master.

PR #477, opened 2018-01-05
Add a snap package (https://snapcraft.io)

popey:
Hi! I put together this PR to add a [snap](https://snapcraft.io) package. I've tested it on Ubuntu 17.10 but it should work just as well on Ubuntu 14.04, 16.04, Linux Mint, Manjaro, Debian, OpenSUSE, Solus, etc.

If you merge and `npm run package:linux` it will create `release/buttercup-desktop_0.25.1_amd64.snap`.

Copy this to a Linux system, [enable snap support](https://docs.snapcraft.io/core/install), then run:
`sudo snap install --dangerous buttercup-desktop_0.25.1_amd64.snap`

Run with `buttercup-desktop` or find it in the launcher.

If you create a developer account and push this to the Snap Store, it can be discovered and installed through GNOME Software and https://snapcraft.io/discover. To create the developer account, [sign up here](https://dashboard.snapcraft.io/dev/account/), then [register the "buttercup-desktop" name](https://dashboard.snapcraft.io/register-snap/?name=buttercup-desktop).

You'll need the `snapcraft` command to push the snap file to the store.
* If you're on a Mac, you can `brew install snapcraft`
* If you're on Linux, it's `sudo snap install --classic snapcraft`

Then you can push Buttercup Desktop out with:
`snapcraft push buttercup-desktop_0.25.1_amd64.snap --release stable`

(You can also push to the Snap Store [programatically from Travis](https://docs.snapcraft.io/build-snaps/ci-integration#using-travis).)

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     },
     "linux": {
       "target": [
+        "snap",
         "rpm",
         "AppImage",
         "deb"
@@ -171,7 +172,7 @@
     "devtron": "^1.4.0",
     "dropbox-fs": "^0.0.5",
     "electron": "^1.7.9",
-    "electron-builder": "^19.45.5",
+    "electron-builder": "^19.53.0",
     "electron-debug": "^1.1.0",
     "electron-devtools-installer": "^2.2.0",
     "electron-installer-dmg": "^0.1.2",


sallar:
Hi @popey. Thanks for this. 
What is the benefit of snap over other versions? I'm not a linux user myself that's why Im asking.

popey:
Snaps have direct exposure to tens of millions of users via GNOME Software with a single package on Ubuntu and other distributions. This makes promotion and discovery a key part of your release strategy, with users able to self-serve install your application rather than hunt it down on an obscure github releases page. We've seen a significant up-tick in users for applications which are promoted through GNOME software. 

Another significant advantage over the legacy packging formats is you having direct control over updates. You can deliver daily builds to adventurous users and stable builds to more conservative users via channels in the store. 

Snaps automatically update, so once you push an update to the store, you can be confident users will get updated promptly, reducing the long tail of versions you get support requests for. We also provide deep analytics of the application user base so you can see what releases users are on, geography etc. 

You only need make one snap package, we deal with ensuring it works on multiple distributions. When new Linux distributions are supported, you don't need to do anything for your application to work there. 

I'm happy to answer any further questions you may have.

perry-mitchell:
This could be tested locally before release :)

popey:
@perry-mitchell do you have a defined set of tests you'd like performed? I'm happy to run through them.

sallar:
@perry-mitchell @popey I can test the release locally :) Should we merge @perry-mitchell?

perry-mitchell:
@sallar I don't see why not! 😊

PR #491, opened 2018-02-03
Add ukrainian locales

vladkampov:
Implementation of Ukrainian localization


sallar:
Thank you for your contribution. Some strings are missing from the translation, could you check those as well?

vladkampov:
Yep, will do that in a bit :)

PR #493, opened 2018-02-03
Added shortcut to lock current archive

### LINKED ISSUE #470: Lock open archive with keyboard shorcut
Hello!

Is it possible to add a keyboard shortcut as `ctrl+l` to lock an open archive?

brooksbecton:
A user can now hit `Ctrl+l` on windows/linux or `Command+l` on Mac to lock the current archive. 

Fixes #470

### DIFF
--- a/src/renderer/components/sidebar-item.js
+++ b/src/renderer/components/sidebar-item.js
@@ -136,6 +136,7 @@ class SidebarItem extends PureComponent {
         ? [
             {
               label: label('lock'),
+              accelerator: `CmdOrCtrl+l`,
               click: this.props.onLockArchive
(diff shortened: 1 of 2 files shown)


src/renderer/components/sidebar-item.js
```diff
@@ -136,6 +136,7 @@ class SidebarItem extends PureComponent {
         ? [
             {
               label: label('lock'),
+              accelerator: `CmdOrCtrl+l`,
```
sallar: make this conditional: `this.props.active ? ... : null`, because if the archive is not active, the shortcut wont work.

PR #494, opened 2018-02-06
Remove Beta notice from README

sallar:


sallar:
@perry-mitchell I agree, and done!

PR #506, opened 2018-02-09
Use channel queue for icon downloads

### LINKED ISSUE #497: Entries list has a performance issue with icons
Since we have enabled icon support, clicking on categories with a long list of entries, causes a lag and delay. I think it's caused by the function that checks for the icon existence. I think that should be done async. Here's a gif of the issue:

lag

(there a 1-2 seconds delay between clicking "Personal" and it actually getting selected. it includes about 200 items)

### LINKED ISSUE #505: The app is taking 100% CPU and freezed
I have upgraded to the latest version, and it has an annoying behavior. When I open the app, let's say I create a new entry, the app is freezing and the laptop's fan is spinning loudly. I can't do anything inside the app for 2 minutes and then it calms down and I can continue what i was doing. It seems it has something to do with the icons fetching.

sallar:
Fixes #497 
Fixes #505

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,8 +3,7 @@
     "react",
     ["env", {
       "targets": {
-        "node": "7.4",
-        "chrome": 56
+        "electron": "1.7"
       },
       "useBuiltIns": true,
       "modules": false,
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,12 +84,20 @@
       }
     },
     "@buttercup/channel-queue": {
-      "version": "0.2.2",
-      "resolved": "https://registry.npmjs.org/@buttercup/channel-queue/-/channel-queue-0.2.2.tgz",
-      "integrity": "sha512-Jk0GMcmgWo2UqZ2b9iI0NoAbqnGQF4J0SCjpnqt7OzRDb60y5W+BJamIsWNBbpKqgc67JL4qhclAFE41BTzwmw==",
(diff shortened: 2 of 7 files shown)


darsain:
Any ETA on when this might ship? The app is pretty much unusable for me atm. Instantly freezes when I click on an item. Not the best first impression :)

sallar:
@darsain This should be actually in version v1.1.0. The issue is still happening? can you give us a bit more info?

darsain:
OS: Windows 10
Version: 1.1.0

I've imported a database form KeePass' `.kdbx` file. Buttercup loaded icons for each item, but some of them I assume failed to resolve and defaulted to this:

image

When I click on an item with this kind of icon, the whole app starts consuming one whole CPU core, freezes, and I have to kill it. Clicking on items with loaded icons doesn't do that.

An example of an item that failed to resolve an icon would be a basic username+password login with URL: https://app.net, which is a website that shut down so I guess it doesn't have favicons available anymore.

sallar:
@darsain so the issue happens when you "click" on the item with no icon? that should not be related to the icons AFAIK. But I'll try to reproduce this. Can you please open an issue for this with the same information so we can track and work on it?

darsain:
Actually I've just tried clicking on various items and it freezes even on the ones with icons. I'll make a new issue.

darsain:
#520

PR #508, opened 2018-02-09
Add random illustrations to empty view

sallar:
Replace the bench image with random illustrations

image

image

image


src/renderer/components/archive/entry.js
```diff
@@ -123,7 +122,7 @@ class Entry extends PureComponent {
         <EmptyView
           caption={t('entry.select-or-create-an-entry')}
           className={styles.emptyView}
-          imageSrc={bench}
+          imageSrc={getRandomIllustration()}
```
perry-mitchell: Would this be safer setting it to state? If a re-render occurs for any reason, the image will change, right?

src/renderer/styles/img/illustrations/3.svg
```diff
@@ -0,0 +1 @@
...-3.91,-3.77c-0.41,-2.08 0.66,-3.75 2.41,-3.74c1.75,0.01 3.5,1.69 3.92,3.76Z" transform="translate(29.2, 63.59)"></path><path fill="#191816" d="M0.38,0c-0.24,0 -0.38,0.23 -0.33,0.5l6.893e-09,4.34007e-08c0.0427821,0.269371 0.259183,0.477606 0.53,0.51c0.23,0 0.37,-0.23 0.32,-0.5l1.33714e-08,7.91848e-08c-0.0446858,-0.264626 -0.254556,-0.47046 -0.52,-0.51Z" transform="translate(31.05, 65.87)"></path><path fill="#FFE500" stroke="#191816" stroke-linecap="round" stroke-linejoin="round" d="M0,3.5723e-08l2.5217e-08,6.16014e-08c0.198893,0.485866 0.648031,0.823679 1.17,0.88l3.3912e-08,-3.65162e-10c0.463892,-0.00499534 0.835902,-0.385104 0.830906,-0.848996c-0.000111337,-0.0103393 -0.000413572,-0.0206757 -0.000906549,-0.0310039" transform="translate(32, 68.6)"></path><path fill="none" stroke="#191816" d="M0,0l5.35,1.52" transform="translate(30.33, 64.59)"></path><path fill="#191816" d="M0.99,1.09647e-08h-0.43c-0.35,0 -0.56,0.33 -0.48,0.74l0.11,0.52l-1.82922e-09,-1.18423e-08c0.0605526,0.392016 0.375888,0.695066 0.77,0.74h0.44c0.34,0 0.55,-0.33 0.47,-0.74l-0.1,-0.51l-1.31727e-08,-7.75299e-08c-0.0669798,-0.39422 -0.383458,-0.698525 -0.78,-0.75Z" transform="translate(32.65, 65.35)"></path></g></svg>
```
perry-mitchell: Should we pick a few more?

sallar: @perry-mitchell yes! what kinds should I find?

src/renderer/components/archive/entry.js
sallar: @perry-mitchell actually this was intentional, so for example if the user sees this view again, the image will change to a new one. I have some mac apps that do this and it's really nice. so you don't have to quit the app to see a new one. what do you think?

src/renderer/styles/img/illustrations/3.svg
sallar: done

src/renderer/components/archive/entry.js
perry-mitchell: I'm happy with it 😊

PR #537, opened 2018-02-21
Remember screen position (#535)

### LINKED ISSUE #535: Remember window position and size
I have 3 monitors. I'd love for buttercup to launch on a specific one with previous size and position, just like every other app does :)

Here is a simple to implement library for this:

https://github.com/mawie81/electron-window-state

ph1p:
**Problem**
Buttercup does not remember the last window position.
This PR should fix it.

I've tested this feature only on macos.

**Issue**
Fixes #535

### DIFF
--- a/src/main/windows.js
+++ b/src/main/windows.js
@@ -15,6 +15,8 @@ export function setupWindows(store) {
   // Intro Screen
   windowManager.setBuildProcedure('main', callback => {
     const [width, height] = config.get('window.size', [950, 700]);
+    const windowPosition = config.get('window.position');
+
     // Create the browser window.
     const win = new BrowserWindow({
       width,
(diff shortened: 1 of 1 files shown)


sallar:
Thanks @ph1p! Can we add a safety check to this? What happens if you have this on your 3rd monitor and you disconnect your screens and re-open buttercup? Will it be moved out of bounds or does electron prevent it from getting moved to far?

ph1p:
Oh good point! I'll check it.

ph1p:
Ok. My first test shows that electron prevents it from getting out of bounds.
- Connected 3 Monitors [_1_] [_2_] [_3_]
- Moved Buttercup from 1 to 3 and disconnected 3.
- Buttercup shows up on first one.
- Connected 3 again and buttercup shows up on this one.

sallar:
@ph1p Awesome!

PR #546, opened 2018-02-22
Make app single instance

### LINKED ISSUE #544: Enforce single instance mode
Buttercup doesn't have a single instance mode enabled. This allows multiple buttercup instances to run, possibly fighting over database access and creating conflicts in settings storage.

Practical case when this might happen: you have buttercup open, and you want to add another .bcup database. You double click on it and instead of it being added to the current instance (or activated when it's already in the list, related issue #543) you get a new buttercup window. And since both instances share the same settings storage, there will definitely be conflicts and confusion.

sallar:
Fixes #544

This change is being tested in Windows. (see issue)

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -12,6 +12,7 @@ import { isWindows } from '../shared/utils/platform';
 import { sleep } from '../shared/utils/promise';
 import { setupActions } from './actions';
 import { setupWindows } from './windows';
+import { getFilePathFromArgv } from './utils/argv';
 
 log.info('Buttercup starting up...');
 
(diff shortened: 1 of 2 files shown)


src/main/app.js
```diff
@@ -71,8 +72,33 @@ app.on('open-file', (e, filePath) => {
...
+    }
+    mainWindow.focus();
+    handleArgvForWindow(mainWindow);
+  } else {
+    windowManager.buildWindowOfType('main', win => handleArgvForWindow(win));
+  }
+});
+
+if (isSecondInstance) {
+  app.quit();
```
perry-mitchell: Hmm, will this create weird behaviour? Would it be possible to focus the first one? I guess that's probably quite difficult..

sallar: Thats what the first few lines do. This is straight out of electron’s docs

sallar: 41c1b875-6b94-426a-b022-9d3f62e27d4d

perry-mitchell: Ah I missed those! Good stuff.

darsain:
I'd also like to note that the time it takes to focus the first instance since the double-click on the file is quite long (a few seconds). Would it be possible to do this check and quit before any unnecessary initialization happens?

sallar:
@darsain we're doing it quite early.. not sure what is happening before it?

Issue #548, opened 2018-02-23
Archive Search


ph1p:
Hi!

I'm using buttercup daily. Often I need a password very quickly and I don't want to click multiple folders to open the entry. There is no solution yet to search globally. So I've written one. I know its not perfect but what do you think?

You can open it via main menu or `cmd/ctrl + f`.

This is currently a `title` search with `findEntriesByProperty('title', '')` from buttercup-core.
It only searchs in the active archive not across all opened archives.

Icons are only placeholders, because I've not found a solution yet to get them.

![search](https://user-images.githubusercontent.com/15351728/36591395-309b5ffe-1892-11e8-957e-88ecd278d544.gif)

sallar:
This is related to #136 and #314.
@perry-mitchell this looks like a good solution to our UI problem?

sallar:
@ph1p can you open a PR so we can discuss there? since this is a code-related discussion and not an actual issue. then we can close this one and mention those are tasks from the PR. You can also make the search to include the other open archives as well. it's what buttercup browser extension does, so its nice to have the same experience.

ph1p:
Ok you can close it :)  I'll sent a PR.

perry-mitchell:
This looks amazing! I love the idea.

For me I think it'd be best to use the `EntryFinder` code in the core, as [used here in the browser extension](https://github.com/buttercup/buttercup-browser-extension/blob/master/source/background/library/archives.js#L175).

Issue #314, opened 2017-07-14
Search does not span folders 


wgregori:
When searching for a password entry the search does not span across folders.  You have to choose the folder then perform the search. Is this the desired result?

perry-mitchell:
This is currently underway when fuzzy search is upgraded. Sallar? 😊

sallar:
@perry-mitchell @wgregori yes this will be done in #136 :) This is duplicate of that so will close :) But thanks for reporting it @wgregori! Cheers

perry-mitchell:
This can be easily added to the desktop application using the `EntryFinder` class in [buttercup core](https://github.com/buttercup/buttercup-core/blob/master/API_WEB.md#EntryFinder).

hugo-cardenas:
What were you thinking about this, UX wise? 

- Should the new across-folders search completely replace per-folder search, just with the same UI?
  - In case yes, as the results would be potentially a mix of entries from all folders, should it visually "deselect" the folder previously selected?
  - Also, when selecting each of the entries in search results, should it mark its parent folder as selected, to show to which folder it belongs?
- Another fully different alternative (as 1Password does, for example) is having instead a dynamically generated "All items" folder, so that actually the search would just work the same as it does now, only that you would select that folder if you would want to search across all (not sure if this plays well with how entries are stored in buttercup-core)

perry-mitchell:
I actually quite like the "All Items" group.. That's a good idea. What do you think @sallar? It would most likely need to be made available in the core, so that every application could easily benefit from it.

I think when searching it should clear the currently selected group..

ph1p:
This is my actual search and I added the `EntryFinder` class. I've only one problem. How can I find the group of an entry? I think this is important, because I can't switch folders without this feature.

![kapture 2017-11-29 at 23 48 18](https://user-images.githubusercontent.com/15351728/33403228-cb313960-d55f-11e7-8856-586cdc682f57.gif)

Don't wonder about: 
- Hidden entry sidebar
- Search in tree view
- entry count next to folders
- titlebar icons
- color (I removed the vibrancy effect to get rid of the mac os bug)
These things are all tests of mine and not for production use or something else.

darsain:
Having to click some "All Items" folder before I can search globally would be very annoying. Not to mention it would clutter my directories with another directory I didn't create and can't get rid of. This can easily be solved with UX.

- Ctrl+F (optional, or just ↓)
- start typing
[...]

Issue #136, opened 2017-01-08
Use fuzzy/deep searching from core


perry-mitchell:
Depends on buttercup-pw/buttercup-core#135

sallar:
@perry-mitchell any updates on this from core?

perry-mitchell:
@sallar Not yet, hopefully soon :)

sallar:
@perry-mitchell Hey man, does the search exist on core now?

perry-mitchell:
Yep! It's used in the browser extension..

PR #549, opened 2018-02-23
Archive Search

ph1p:
[Copy from #548]

Hi!

I'm using buttercup daily. Often I need a password very quickly and I don't want to click multiple folders to open the entry. There is no solution yet to search globally. So I've written one. I know its not perfect but what do you think?

You can open it via main menu or cmd/ctrl + f.

This is currently a title search with findEntriesByProperty('title', '') from buttercup-core.
It only searchs in the active archive not across all opened archives.

Icons are only placeholders, because I've not found a solution yet to display them.

--------

**New gif** 

search

> sallar commented an hour ago • 
> This is related to #136 and #314.
> @perry-mitchell this looks like a good solution to our UI problem?
> we can also rename the current "search..." on top of the entries list to "filter..." (so it works on the current list)
> 
> 
> sallar commented 40 minutes ago
> @ph1p can you open a PR so we can discuss there? since this is a code-related discussion and not an actual issue. then we can close this one and mention those are tasks from the PR. You can also make the search to include the other open archives as well. it's what buttercup browser extension does, so its nice to have the same experience.
> 
> 
> ph1p commented 25 minutes ago
> Ok you can close it :) I'll sent a PR.
> 
> perry-mitchell commented 20 minutes ago
> This looks amazing! I love the idea.
> 
> For me I think it'd be best to use the EntryFinder code in the core, as used here in the browser extension.


src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,279 @@
...
+  closeSearch() {
+    this.setState({
+      visible: false
+    });
+  }
+
+  changeInput(e) {
+    const value = e.target.value;
+    const entries = value
+      ? this.state.archive.findEntriesByProperty('title', value)
```
perry-mitchell: Please see [my earlier comment](https://github.com/buttercup/buttercup-desktop/issues/548#issuecomment-367999671). Would be great to use `EntryFinder` here instead.

sallar:
@ph1p what happened? :D

PR #550, opened 2018-02-23
Feature/archive search

### LINKED ISSUE #136: Use fuzzy/deep searching from core

### LINKED ISSUE #314: Search does not span folders

ph1p:
Fixes #136 
Fixes #314

[Copy from #548]

Hi!

I'm using buttercup daily. Often I need a password very quickly and I don't want to click multiple folders to open the entry. There is no solution yet to search globally. So I've written one. I know its not perfect but what do you think?

You can open it via main menu or cmd/ctrl + f.

This is currently a title search with findEntriesByProperty('title', '') from buttercup-core.
It only searchs in the active archive not across all opened archives.

Icons are only placeholders, because I've not found a solution yet to display them.

--------

**New gif** 

search

> sallar commented an hour ago • 
> This is related to #136 and #314.
> @perry-mitchell this looks like a good solution to our UI problem?
> we can also rename the current "search..." on top of the entries list to "filter..." (so it works on the current list)
> 
> 
> sallar commented 40 minutes ago
> @ph1p can you open a PR so we can discuss there? since this is a code-related discussion and not an actual issue. then we can close this one and mention those are tasks from the PR. You can also make the search to include the other open archives as well. it's what buttercup browser extension does, so its nice to have the same experience.
> 
> 
> ph1p commented 25 minutes ago
> Ok you can close it :) I'll sent a PR.
> 
> perry-mitchell commented 20 minutes ago
> This looks amazing! I love the idea.
> 
> For me I think it'd be best to use the EntryFinder code in the core, as used here in the browser extension.

### DIFF
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -48,6 +48,7 @@ export const setupMenu = store => {
           type: 'separator'
         },
         {}, // Import menu will be injected here
+        {}, // Import search menu seperator
         {}, // Import search menu
         {
           type: 'separator'
@@ -282,9 +283,6 @@ export const setupMenu = store => {
                 accelerator: 'CmdOrCtrl+F',
                 click: (item, focusedWindow) => openSearch(focusedWindow)
               };
-            }
-            if (i === 7) {
-              return { type: 'separator' };
             }
             return sub;
           })
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -8,7 +8,7 @@ import {
 import { setSetting } from '../shared/actions/settings';
 import { ImportTypeInfo } from '../shared/buttercup/types';
 import { openFile, openFileForImporting, newFile } from './lib/files';
-import { openSearch } from './lib/archive-search';
+import { toggleArchiveSearch } from './lib/archive-search';
 import { getWindowManager } from './lib/window-manager';
 import { checkForUpdates } from './lib/updater';
 import { getMainWindow } from './utils/window';
@@ -48,8 +48,14 @@ export const setupMenu = store => {
           type: 'separator'
         },
         {}, // Import menu will be injected here
-        {}, // Import search menu seperator
-        {}, // Import search menu
+        {
+          type: 'separator'
+        },
+        {
+          label: label('archive.search'),
+          accelerator: 'CmdOrCtrl+F',
+          click: (item, focusedWindow) => toggleArchiveSearch(focusedWindow)
+        },
         {
           type: 'separator'
         },
@@ -271,17 +277,6 @@ export const setupMenu = store => {
                           }
                         ]
                 }))
-              };
-            }
-            // Archive Search
-            if (i === 5) {
-              return { type: 'separator' };
-            }
-            if (i === 6) {
-              return {
-                label: label('archive.search'),
-                accelerator: 'CmdOrCtrl+F',
-                click: (item, focusedWindow) => openSearch(focusedWindow)
               };
             }
             return sub;
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -8,6 +8,7 @@ import {
 import { setSetting } from '../shared/actions/settings';
 import { ImportTypeInfo } from '../shared/buttercup/types';
 import { openFile, openFileForImporting, newFile } from './lib/files';
+import { openSearch } from './lib/archive-search';
 import { getWindowManager } from './lib/window-manager';
 import { checkForUpdates } from './lib/updater';
 import { getMainWindow } from './utils/window';
@@ -47,6 +48,14 @@ export const setupMenu = store => {
           type: 'separator'
         },
         {}, // Import menu will be injected here
+        {
+          type: 'separator'
+        },
+        {
+          label: label('archive.search'),
+          accelerator: 'CmdOrCtrl+F',
+          click: (item, focusedWindow) => openSearch(focusedWindow)
+        },
         {
           type: 'separator'
         },
--- a/src/main/menu.js
+++ b/src/main/menu.js
@@ -48,14 +48,7 @@ export const setupMenu = store => {
           type: 'separator'
         },
         {}, // Import menu will be injected here
-        {
-          type: 'separator'
-        },
-        {
-          label: label('archive.search'),
-          accelerator: 'CmdOrCtrl+F',
-          click: (item, focusedWindow) => openSearch(focusedWindow)
-        },
+        {}, // Import search menu
         {
           type: 'separator'
         },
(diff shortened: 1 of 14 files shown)


src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,279 @@
...
+  closeSearch() {
+    this.setState({
+      visible: false
+    });
+  }
+
+  changeInput(e) {
+    const value = e.target.value;
+    const entries = value
+      ? this.state.archive.findEntriesByProperty('title', value)
```
perry-mitchell: Copied from the last PR:

> Please see [my earlier comment](https://github.com/buttercup/buttercup-desktop/issues/548#issuecomment-367999671). Would be great to use `EntryFinder` here instead.

src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,279 @@
...
+      : [];
+    const groups = value ? this.state.archive.findGroupsByTitle(value) : [];
```
perry-mitchell: Do we need to search groups? I can't really see there being that many groups that they'd need to be included in search results. What's your opinion @sallar @ph1p ?

src/renderer/components/archive/archive-search.js
ph1p: Sorry about the second PR. Do not know if you can delete my last pull request. I've committed sensitive data 😰

ph1p: I'll try to add the EntryFinde ;)

src/renderer/components/archive/archive-search.js
ph1p: @perry-mitchell I asked a colleague of mine for his opinion on this PR. He said it would be nice to see folders. So I added them. For me only entries are important.

src/renderer/components/archive/archive-search.js
perry-mitchell: @ph1p I've contacted Github to have the PR removed.

ph1p: Thanks!

src/renderer/components/archive/archive-search.js
perry-mitchell: Perhaps @sallar has a point on this? :smile:

src/renderer/components/archive/archive-search.js
perry-mitchell: @ph1p No problem.

If it does get deleted then the example for `EntryFinder` is [here](https://github.com/buttercup/buttercup-browser-extension/blob/master/source/background/library/archives.js#L186) if you need it.

darsain:
If it'll eventually be able to search among all open databases, make it configurable please. Searching all of them would mean unwanted noise and inconsistency in search results for me.

Don't make the search box jump up and down trying to center it vertically as the results are being populated. It's chaotic. Position it statically towards the top and expand results below.

And I definitely prefer the original search box:

image

Over the current hint like design:

image

The search results really don't seem like hints, so why put them in a hint box? :)

perry-mitchell:
> If it'll eventually be able to search among all open databases, make it configurable please. Searching all of them would mean unwanted noise and inconsistency in search results for me.

I somewhat agree here. I don't think this will be an issue for this PR, and could definitely be discussed as a separate issue I think.

> Don't make the search box jump up and down trying to center it vertically as the results are being populated. It's chaotic. Position it statically towards the top and expand results below.

I like this idea.. The original box did look a bit cleaner... ;)

ph1p:
@perry-mitchell @darsain I'll do my best.

perry-mitchell:
@ph1p I think you've already done an amazing job! These are but suggestions. Your first approach was just really net imho. Keep up the great work - we can't wait to merge this as a feature!

ph1p:
thanks @perry-mitchell :) My code uses the `EntryFinder` now. I've removed the groups.

ph1p:
And finally I've added the icons :)

ph1p:
Now the EntryFinder searchs across all opened archives and shows the archive title in your search results.

src/renderer/components/archive/archive-search.js
perry-mitchell: @ph1p The old branch history has been deleted (just received confirmation from GH) 👍

src/main/menu.js
```diff
@@ -47,6 +48,8 @@ export const setupMenu = store => {
           type: 'separator'
         },
         {}, // Import menu will be injected here
+        {}, // Import search menu seperator
```
sallar: You can just directly put the menu items here.  You don't need to update them later. That is for the cases where that menu is dynamic

src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }
+`;
+
+const EntryData = styled.div`
+  display: inline-block;
+  margin: -5px 0 0 0;
+`;
+
+const EntryFolder = styled('p')`
```
sallar: `styled('p')` -> `styled.p`

src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,251 @@
...
+            }}
+            onChange={this.changeInput}
+            value={this.state.searchTerm}
+            placeholder={t('archive-search.searchterm')}
+            type="text"
+            entries={this.state.entries}
+            searchTerm={this.state.searchTerm}
+          />
+
+          {entries.length > 0 ? (
```
sallar: Instead of this ternary operator, you can use `<Choose>` and `<When>`, we have been using them in other files
https://www.npmjs.com/package/jsx-control-statements

src/renderer/components/archive/archive-search.js
```diff
@@ -0,0 +1,251 @@
...
+            <EntryList flexAuto>
+              <Scrollbars autoHeight autoHeightMax={300}>
+                {entries.map(({ entry, sourceID, groupID, icon }, index) => (
+                  <ListItem
+                    key={index}
+                    onClick={() => {
+                      switchArchive(sourceID);
```
sallar: These should be in a redux action, and here you would only trigger that action instead of calling them individually.

src/renderer/components/workspace.js
```diff
@@ -43,6 +44,7 @@ const Workspace = ({
                 <NoArchiveSelected />
               </When>
               <Otherwise>
+                <ArchiveSearch />
```
sallar: Instead of listening to ipc actions direction in the component, use `getUIState` and `setUIState` action and selector to toggle between search states, and pass that value here so you can display the component based on that value: `{searchVisible && <ArchiveSearch />` for example. And in `renderer/index.js` you would listen to the ipc:

```js
ipc.on('open-archive-search', () => {
  store.dispatch(setUIState('searchVisible', true));
});
```

ph1p: Ok I've pushed my update :) Now you can control the result with arrow up/down and enter key.

ph1p:
Yeah 👍

PR #551, opened 2018-02-24
Feature/search improvements

sallar:


src/renderer/components/archive/archive-search.js
```diff
@@ -184,6 +184,10 @@ class ArchiveSearch extends PureComponent {
         this.openEntry(result.sourceID, result.entry);
       }
     }
+
+    if (e.keyCode === 27) {
+      this.closeSearch();
```
perry-mitchell: @sallar does it also close on unfocus? Would be good if it was also possible via mouse.

sallar: Yeah if you click out or press esc it closes the box

perry-mitchell: 👍

PR #552, opened 2018-02-25
Feature/tray icon

### LINKED ISSUE #444: Support tray-icon/background mode in Windows/Linux
Hello,
There is possibility to minimise in windows toolbar ? Like internet access icon

ph1p:
Hi!

I've created a tray icon. Now we've the ability to switch from dock to tray icon and back.

Fixes #444


src/main/menu.js
```diff
@@ -364,5 +377,27 @@ export const setupMenu = store => {
...
+    app.dock.show();
+
+    setTimeout(() => {
+      if (app.dock.isVisible()) {
+        if (tray) {
+          tray.destroy();
+        }
+        tray = null;
+      }
+    }, 300);
```
perry-mitchell: What's with the timeout? 😊

src/main/lib/utils.js
```diff
@@ -1,8 +1,14 @@
 import path from 'path';
 
-export function getPathToFile(fileName) {
+export function getPathToFile(fileName, disableFilePath = false) {
```
perry-mitchell: I would suggest using a whole new function for this, rather than having a cryptic boolean flag.. What about `getURIPathToFile` and `getPathToFile`?

Changing them would be easy due to their infrequent use.

perry-mitchell:
@ph1p So could you walk us through the behaviour?
 - What is the default behaviour?
 - Is the tray icon switchable?
 - What do you mean by the following:

> In development all is good, but in production something went wrong and I don't know why :/

perry-mitchell:
@sallar If we choose to move this on through, what do you think about altering the tray behaviour on mac to match windows/linux? Currently mac is the odd one out - if we have a switch, having the tray icon _disabled_ I would expect the mac application to fully quit when closing the last window.. but currently only windows & linux do that.

src/main/menu.js
ph1p: app.dock.show() is asynchronous. When I disable the tray icon and enable the dock icon at the same time, the app crashes.

ph1p:
First of all, this is only an idea and maybe it is always there and you cannot disable it. I started this because of #444.

---

_What is the default behaviour?_
The tray icon is disabled

_When its active_
screen shot 2018-02-25 at 19 06 46 2

---

_Is the tray icon switchable?_
Jep. You can use the `setImage` method.

---

_What do you mean by the following: [...]_
I'm able to see the tray icon in development mode, but when I run `npm run build && npm run pack:mac` the app shows this and an error window shows up and closes itself immediately.

screen shot 2018-02-25 at 19 04 14

src/main/menu.js
perry-mitchell: Can you use `0` as the value? Having arbitrary values (300) is not so stable and the necessary value could differ from computer to computer.

src/main/lib/utils.js
```diff
@@ -1,14 +1,15 @@
 import path from 'path';
 
-export function getPathToFile(fileName, disableFilePath = false) {
+export function getPathToFile(fileName) {
   if (process.env.NODE_ENV === 'development') {
-    return `${disableFilePath ? '/' : 'file://'}${path.resolve(
-      __dirname,
-      `../../../app/${fileName}`
-    )}`;
+    return `file://${path.resolve(__dirname, `../../../app/${fileName}`)}`;
```
perry-mitchell: Sorry to be pedantic! But `getPathToFile` should return something like `/path/to/file` and `getURIPathToFile` should return the `file:///path/to/file` URL.

To make it simpler, could could make `getPathToFile`, and then:

```javascript
function getURIPathToFile(fileName) {
  return `file://${getPathToFile(fileName)}`;
}
```

(Most likely)

ph1p: Sure :)

sallar:
@ph1p how does the icon look on mac (light/dark), windows and linux?

ph1p:
@sallar Currently I'm working on a good user experience and test out various things. 

**This is how it looks like on mac and windows**
screen shot 2018-02-26 at 12 23 55
screen shot 2018-02-26 at 12 24 56
screen shot 2018-02-26 at 12 25 36
screen shot 2018-02-26 at 12 25 22

src/main/lib/utils.js
```diff
@@ -1,8 +1,12 @@
 import path from 'path';
 
 export function getPathToFile(fileName) {
```
perry-mitchell: Sorry - this should be `getURIPathToFile`, as it uses the `file:/` uri.

src/main/lib/utils.js
```diff
@@ -1,8 +1,12 @@
...
+  return `file://${getURIPathToFile(fileName)}`;
+}
+
+export function getURIPathToFile(fileName) {
```
perry-mitchell: This should be `getPathToFile` as it returns a regular file path.

ph1p:
@sallar I've added an .ico file for windows and on mac two files, one 16x16 and 32x32 (retina). I've tried different sizes and its always a little bit blurry. It would be nice, if you can try it by yourself or provide some icons.

src/main/lib/utils.js
ph1p: k

src/main/lib/utils.js
ph1p: k

sallar:
Actually the quality is ok on mac:

image

There is a problem though, when I right click on the menubar Item, it takes a few tries to get rid of the popup menu... Dont know why

src/main/menu.js
```diff
@@ -15,11 +15,98 @@ import { getMainWindow } from './utils/window';
...
 import electronContextMenu from 'electron-context-menu';
+import { getPathToFile } from './lib/utils';
 
 electronContextMenu();
 
 const label = (key, options) => i18n.t(`app-menu.${key}`, options);
 
+let tray = null;
+
+const reopenMainWindow = store => {
```
sallar: Why are you passing `store` to this function but not using it?

perry-mitchell:
Hmm.. not sure how much I love the icon on windows (the lock looks amazing on Mac).. The green lock icon by itself seems a bit on the dull side. @sallar what would you think about using our actual icon for windows?

sallar:
@perry-mitchell @ph1p yeah I agree. Can we give the actual square icon a try on windows and linux? I think it'll look good on small sizes too

perry-mitchell:
>  actual square icon

@sallar Did you mean the rounded corners one, or the actual square icon? :smile:

ph1p:
@sallar @perry-mitchell you're right ^^ looks a little bit out of place. I'll change it and improve the context menu behavior.

I'll try this: screen shot 2018-02-27 at 14 13 14

perry-mitchell:
@ph1p That looks much better imo.. the rounded-corners icon. Good stuff.

So, to confirm that the behaviour is as intended, could you ensure that this is how it's running now? So at least we're all on the same page (correct me if I've made a mistake):

## On windows + linux

 * If the tray icon is **disabled**:
   * Closing the window exits the application
 * If the tray icon is **enabled**:
   * Closing the window closes only the window selected (the tray icon remains)

## On mac

 * Closing the window leaves the app active in the background (existing behaviour)

perry-mitchell:
Also, should we add another button to the menu to simply open Buttercup? @sallar @ph1p ?

sallar:
@perry-mitchell clicking on the icon opens buttercup, right clicking shows brings up that menu

sallar:
@ph1p I pushed a change to your branch :D sorry

ph1p:
@sallar no problem :)

@perry-mitchell this is correct, after my last commit! I've tested it only on mac and windows.

ph1p:
@perry-mitchell I think the additional menu item to open buttercup is a good idea.

Maybe we should add the menu on both clicks.

This is an example how others apps handle these clicks. All of them are very different, but most of them open the menu on left click. I thought, it would be nice to open buttercup as fast as possible, but maybe it confuses some people.

kapture 2018-02-28 at 0 43 12
green = right click

perry-mitchell:
@ph1p I personally agree on that point.. I like how the Docker menu works -> both clicks open the menu. Very simple! @sallar?

sallar:
@ph1p is this PR ready?

sallar:
There is a bug. When a save is going on, and you choose "Quit" from the menu, only the window closes, but it doesnt quit the app because of the new setting you added in `window-all-closed` event... But if no save is going on, then you can quit without a problem.

darsain:
Just tested this out on Win 10:

1. Clicking on the icon doesn't open the app, but shows the context menu, the same thing that happens upon right-click. This is pretty useless and non-standard.
I see that this is actually argued *for* above? Why? :( I'd never want both buttons to do the same thing. Left click is the easiest shortcut to the most used and intuitive action on that menu - opening the app. I can't image better UX here. I confess I've never used mac as a primary OS before but it baffles my mind if this is standard there. If the app has GUI, I always expect clicking on its tray icon to show the GUI.

1. Had to look for the toggle a bit, since none calls the tray "menubar".

perry-mitchell:
> This is pretty useless and non-standard

@darsain I disagree - almost all of my tray applications do this (the same for both clicks), including ownCloud, Docker, Keybase and F.lux. I think it's much easier so you don't have to remember specific behaviour for _every_ tray application. Seems the other apps have the same idea.

> Had to look for the toggle a bit, since none calls the tray "menubar".

This is pretty minor imo, but we're considering forcing the tray icon feature for everyone. Not sure why we might want 2 modes of operation for the future. We're considering updating this, but haven't decided yet.

PR #592, opened 2018-04-14
Fix Tray Issues

### LINKED ISSUE #573: Multiple icons after update
I now have multiple Buttercup icons after updating from 1.5.1 to 1.5.2:

image

### LINKED ISSUE #581: Tray error on every load
On ubuntu 16.04, I get this error every time I open buttercup:

image

I have the tray icon enabled.

sallar:
Fixes #581 
Fixes #573


src/main/app.js
```diff
@@ -100,7 +100,8 @@ const isSecondInstance = app.makeSingleInstance(argv => {
 });
 
 if (isSecondInstance) {
-  app.quit();
+  log.info('Second instance opened, attempting to exit now.');
+  app.exit();
```
perry-mitchell: What is the behaviour of `isSecondInstance` in this case:

 1. Version 10.0.0 is running
 2. Version 10.1.0 is started

Does version 10.0.0 close? Because I'd imagine it should, especially in the case of using AppImage builds..

sallar: Second instance is just basically the second running instance regardless of version. But I imagine we can configure the behavior by manually checking the versions and alerting the user or something? But if so, lets create an issue for it and fix it separately

perry-mitchell: > regardless of version

Ok, so that means that electron doesn't know which is the "second instance"? Unfortunately it might be quite critical: If the user is trying to open a new instance, and they cannot due to it being closed constantly, then usability isn't so great..

Wouldn't it be better to force-close the first instance, and always keep the _newly opened instance_ active?

perry-mitchell: To move this forward, I created #597 to take care of the execution/closing.

PR #614, opened 2018-05-26
New Italian strings + some edits

fortym2:
Hi, I've added new strings and modified some already translated words, mainly because they were ambiguous in the context.
For example, "cancel" is better translated to "annulla" than "cancella" (that means "delete").


sallar:
Thanks @fortym2

Issue #634, opened 2018-07-28
prevent loss of unsaved new entry


leongrdic:
when adding a new entry, I often find myself forgetting to actually save it, and later switching to another (already existing) entry to copy a different password. this way i lose the new, unsaved entry.
a possible solution to avoid this problem would be a dialog that appears when trying to open another entry while the current one isn't saved, saying "you didn't save your entry, you will lose the data if you don't save it now" or something similar.
same thing goes for editing existing entries!

perry-mitchell:
I like this idea. Editing could almost grey-out/disable the other items to prevent this, or a dialog like you suggested.

What do you think @sallar ?

sallar:
@perry-mitchell fortunately we are using redux-form so we *can* persist the data and show it again later, but IMO showing a dialog to confirm navigating away is a better option. Because the goal here would be to prevent the user from "forgetting" right?

sallar:
This issue isn't fixed in master. So won't be included in release notes.

leongrdic:
okay, i opened the new issue.
i just noticed that closing the buttercup window (app) while adding a new entry doesnt show any warning or a chance to prevent the user from losing the new unsaved entry. should i also open a new issue for this?

ldexterldesign:
> I just noticed that closing the buttercup window (app) while adding a new entry doesn't show any warning or a chance to prevent the user from losing the new unsaved entry. should i also open a new issue for this?

Yes, I also noticed this today

IMO *any* entry change that's unsaved should trigger a save prompt if the user events outside the entry column/pane:

<img width="554" alt="screenshot 2019-01-21 at 16 13 04" src="https://user-images.githubusercontent.com/1715809/51482733-87601900-1d97-11e9-88dc-b1c1a0b0baba.png">

One new issue should suffice

Cheers

PR #666, opened 2018-10-14
feat: prevent loss of an unsaved entry.

antoine-amara:
When creating a new entry, if a user click on an another entry to view
it, ask to him to confirm his action to prevent a data loss.

Update the onClick event on a list element (entries-list) to
be able to check the state of the app and display the dialog window
if it is needed. Create a redux selector to retrieve the mode field
to be able to check into the list if a new entry is opened.

Fixes: https://github.com/buttercup/buttercup-desktop/issues/634

### DIFF
--- a/src/shared/actions/entries.js
+++ b/src/shared/actions/entries.js
@@ -2,7 +2,13 @@ import { createAction } from 'redux-actions';
 import * as entryTools from '../buttercup/entries';
 import { showDialog, showConfirmDialog } from '../../renderer/system/dialog';
 import { getQueue } from '../../renderer/system/queue';
-import { getCurrentGroupId, getCurrentArchiveId } from '../selectors';
+import {
+  getCurrentGroupId,
+  getCurrentArchiveId,
+  getCurrentEntry,
+  getCurrentEntryMode,
+  getEditFormElement
+} from '../selectors';
 import i18n from '../i18n';
 import { EntryFinder } from 'buttercup/dist/buttercup-web.min';
 import { getSharedArchiveManager } from '../buttercup/archive';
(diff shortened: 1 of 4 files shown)


locales/fa/base.json
```diff
@@ -150,7 +150,8 @@
     "custom-fields-label-empty-info": "شما افزودن عنوان به یکی از فیلدهای شخصی را فراموش کردید.",
     "are-you-sure-question": "برای این کار اطمینان دارید؟",
     "displaying-entries": "در حال نمایش {{count}} سند",
-    "no-value": "بدون محتوی"
+    "no-value": "بدون محتوی",
+    "quit-unsave-entry": "سوف تفقد دخولك ، هل أنت متأكد؟"
```
sallar: Please remove the new translations except English. This is up to collaborators to add the correct translations. I see that you have used Google Translate for these which is not correct :)

src/renderer/components/archive/entries-list.js
```diff
@@ -54,13 +56,25 @@ ListItemContent.propTypes = {
...
+}) => (
   <ListWrapper>
     <For each="entry" of={entries}>
       <ListItem
         key={entry.id}
         active={currentEntry && entry.id === currentEntry.id}
-        onClick={() => onSelectEntry(entry.id)}
+        onClick={() =>
+          !currentEntry && currentEntryMode === 'new'
+            ? showConfirmDialog(
```
sallar: This should be handled in the redux actions file. Search the project for `showConfirmDialog` usages to see how it's done in other places. It should not be handled inside the presentation components.

antoine-amara:
Thanks for the review @sallar , I made the changes and I think I found the way to handle the check into the redux action. About translations, I keep english and french (because I'm french).

src/shared/actions/entries.js
```diff
@@ -20,7 +20,27 @@ import {
...
 import { loadGroup } from '../../shared/actions/groups';
 
-export const selectEntry = createAction(ENTRIES_SELECTED);
+export const selectEntry = (
+  entryId,
+  currentEntry,
+  currentEntryMode
+) => async dispatch => {
+  try {
+    !currentEntry && currentEntryMode === 'new'
```
sallar: You can get `currentEntry` and `currentEntryMode` here directly using `getState` and the selector functions, no need to pass them in from the UI component.

antoine-amara:
Hello @sallar , yes I will push the changes soon, I'm on it.

sallar:
@antoine-amara the tests are failing :) please have a look:
https://travis-ci.org/buttercup/buttercup-desktop/builds/458136631?utm_source=github_status&utm_medium=notification

antoine-amara:
Hello, I see, sorry for that, I forget to delete an unused import. You prefer to have a new commit for the fix or I have to ammend my first commit ?

Issue #62, opened 2016-02-05
Minimum master password strength


perry-mitchell:
Master passwords should be of a certain strength, without exception. This is more to stop stupidly simple passwords rather than enforcing crazy-complex ones.

perry-mitchell:
This depends on #1

antoine-amara:
I can propose a solution to this issue, I finish my commit and send a PR asap to show you.

PR #669, opened 2018-10-21
feat: check master password strength when creating a new archive.

antoine-amara:
to be more secure, check the master password strength when creating an archive
and return an error if the password doesn't follow the security
recommendations.

for that, follow the main recommendations from the OWASP association.
the master password should now contain:
  - one uppercase letter.
  - one lowercase letter.
  - one number.
  - one special character.

Fixes: https://github.com/buttercup/buttercup-desktop/issues/62
Refs: https://www.owasp.org/index.php/Authentication_Cheat_Sheet#Implement_Proper_Password_Strength_Controls

### DIFF
--- a/locales/en/base.json
+++ b/locales/en/base.json
@@ -210,7 +210,8 @@
     "password": "Password",
     "confirm-password": "Confirm Password",
     "confirm": "Confirm",
-    "nevermind": "Nevermind"
+    "nevermind": "Nevermind",
+    "not-strong-enought": "Your password must contain at least: 1 uppercase letter, 1 number, 1 special character and contain 10 characters."
   },
   "search": {
     "search": "Search"
--- a/locales/fr/base.json
+++ b/locales/fr/base.json
@@ -210,7 +210,8 @@
     "password": "Mot de passe",
     "confirm-password": "Confirmation du mot de passe",
     "confirm": "Confirmer",
-    "nevermind": "Pas grave…"
+    "nevermind": "Pas grave…",
+    "not-strong-enought": "Votre mot de passe doit au moins contenir: 1 majuscule, 1 caractère alphanumérique, 1 caractère spécial et doit faire aux moins 10 caractères."
   },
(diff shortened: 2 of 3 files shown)


sallar:
@antoine-amara Thanks for the PR. However we were thinking about a way more similar to the password strength meter that is already inside the "Create new Entry" page (a progress bar?). What do you think?

perry-mitchell:
Also, I don't think having any custom function to do this is a good idea. We should use a component to do this which we can use across our entire platform. I would also prefer to use one that's designed for this.

We use [this library](https://www.npmjs.com/package/owasp-password-strength-test) in our latest project @sallar - Maybe some wrapper around it could work?

perry-mitchell:
I do believe that OWASP guidelines are the way to go at least.. I think that's safe.

sallar:
@perry-mitchell this looks good! should we also change the existing strength meter to use this one? We currently use this: https://github.com/dropbox/zxcvbn

antoine-amara:
Hello, thanks for the review.

@sallar , I think this is a good idea to reuse the strength meter for the master password. Maybe we have to add a label/message to inform the user has to enter a strength password to continue.

@perry-mitchell , the package [owasp-password-strength-test](https://github.com/nowsecure/owasp-password-strength-test) look good and I can use it for password validation.

I can make these change and update the PR soon:
* reuse the strength meter into the master password form.
* create a component to wrap the password form input and valid it (with owasp-password-strength-test package)

sallar:
@antoine-amara I'm so sorry that I need to close this PR without merging it. With the way the password dialog was designed before, we couldn't add a "strength meter" to the dialog. So I had to re-code the dialog in #744 and I added a strength meter there. 

I hope this doesn't put you off and you will contribute to Buttercup again. Cheers.

antoine-amara:
Ok @sallar thanks. I will continue to contribute when I have more time, I like this project. See you 🙂

PR #721, opened 2018-12-12
fix: tolerate invalid RegExp patterns in archive search

aleclarson:
To reproduce:
- Unlock an archive
- Use cmd+F
- Type "s["

__Buttercup v1.12.2__

### DIFF
--- a/src/renderer/components/archive/archive-search.js
+++ b/src/renderer/components/archive/archive-search.js
@@ -172,13 +172,19 @@ class ArchiveSearch extends PureComponent {
   };
 
   /**
-   * Wrap word with mark tag
-   * @param {string} word text to wrap
+   * Wrap text with <mark> tag
+   * @param {string} text
    * @memberof ArchiveSearch
    */
-  highlightSearchResult = word => {
-    const regex = new RegExp('(' + this.state.searchTerm + ')', 'ig');
-    return word.replace(regex, '<mark>$1</mark>');
+  highlightSearchResult = text => {
+    const startIndex = text.toLowerCase().indexOf(this.state.searchTerm.toLowerCase());
+    if (startIndex >= 0) {
+      const endIndex = startIndex + this.state.searchTerm.length;
+      return text.substring(0, startIndex) +
+        '<mark>' + text.substring(startIndex, endIndex) + '</mark>' +
+        text.substring(endIndex);
+    }
+    return text;
   };
 
   /**


src/renderer/components/archive/archive-search.js
```diff
@@ -177,8 +177,11 @@ class ArchiveSearch extends PureComponent {
    * @memberof ArchiveSearch
    */
   highlightSearchResult = word => {
-    const regex = new RegExp('(' + this.state.searchTerm + ')', 'ig');
-    return word.replace(regex, '<mark>$1</mark>');
+    try {
+      const regex = new RegExp('(' + this.state.searchTerm + ')', 'ig');
+      return word.replace(regex, '<mark>$1</mark>');
+    } catch(e) {}
+    return word;
```
perry-mitchell: Good idea! Though I'd recommend making this simpler now that we're looking at it. Using a regex is obviously going to cause issues when special chars are used, so we should opt for plain text.

Simply finding the index and replacing the word from that point only should do the trick What about something like this:

```javascript
const index = word.toLowerCase().indexOf(this.state.searchTerm.toLowerCase());
if (index >= 0) {
  return `${word.substring(0, index)}<mark>${this.state.searchTerm}</mark>${word.substring(index + this.state.searchTerm.length)}`;
}
return word;
```

perry-mitchell: I haven't tested if this works, but it'd do away with the complexity of regular expressions.

aleclarson: How about escaping any `RegExp` characters in the search term using [escape-string-regexp](https://www.npmjs.com/package/escape-string-regexp)?

aleclarson: We can avoid the dependency by copy-pasting.

```js

  highlightSearchResult = word => {
    const searchTerm = this.state.searchTerm.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&');
    const regex = new RegExp('(' + searchTerm + ')', 'ig');
    return word.replace(regex, '<mark>$1</mark>');
  };
```

aleclarson: Pushed. 😄

perry-mitchell: Sorry I didn't reply sooner. It's still not as robust as what I suggested, I'm afraid. There's no need for regex at all in this situation. It's a simple result of highlighting the first instance of the search term, which my example does with only string splitting.

sallar:
@perry-mitchell this looks good to me, are you happy with it too mate?

PR #727, opened 2018-12-27
Add devtools warning

### LINKED ISSUE #716: Add security warning in console
A security warning should be added to the devtools console:

```
Attention!

Running commands here can put your vault's contents at great risk - you should never run code here that was given to you or that you found online. If you do not understand what you're doing, we recommend that you close this console.

- Buttercup
```

The warning should be localised and in a large stylised font.

LukasWerfel:
closes #716


perry-mitchell:
This looks good from my point of view.. @sallar ?

sallar:
Looks good to me for now too!

perry-mitchell:
@LukasWerfel Sorry for the delay on this.

Any chance you could up the font-size and do some formatting to maybe make it read a bit more strongly?

For instance, Paypal has the following:

image

It might be good to have something similar for us - ie. a larger header and then some warning text. Maybe no need for the "Buttercup" part.

sallar:
I agree with @perry-mitchell a coloured output would be much nicer

src/renderer/containers/root.js
```diff
@@ -8,6 +8,13 @@ export default class Root extends PureComponent {
     store: PropTypes.object.isRequired
   };
 
+  componentDidMount() {
+    console.warn(`Attention!
```
sallar: Also can we add a check to show this only in production? What do you think @perry-mitchell? 

```js
if (process.env.NODE_ENV === "production") {
```

perry-mitchell: Sure! I don’t think it’s the worst if it’s everywhere.. but no harm either way really.

sallar: @perry-mitchell you're right. its fine if it's everywhere

LukasWerfel:
Sure, will also localize it!

LukasWerfel:
@sallar @perry-mitchell 
[Done](https://imgur.com/a/9Ru5kSD)!

Though it`s still lacking translation. I`d do that in a subsequent PR since I first have to get familiar with the intl library.

perry-mitchell:
@LukasWerfel Hey that's super cool!

I still wouldn't worry about localisation at this early stage.. I think this is pretty great as is.

PR #729, opened 2018-12-29
Fix error message for webdav connection error

### LINKED ISSUE #728: Error dialog doesnt show url correctly in webdav.
Steps to reproduce:
1. Add a new webdav vault
1. Enter `https://test.com` and some password
1. Click connect

It will show
> Connection to `https:&#x2F;&#x2F;test.com&#x2F;` failed. Please check your credentials and try again.

The reason for that is that the error message gets translated [here](https://github.com/buttercup/buttercup-desktop/blob/master/src/renderer/components/file-manager/sources/webdav.js#L98-L100). 

And according to [react-i18next docs](https://www.i18next.com/translation-function/interpolation#unescape) we need to escape this to show it properly.

LukasWerfel:
Unescaped it according to [react-i18next docs](https://www.i18next.com/translation-function/interpolation#unescape).

The docs also mention
> Warning: If you toggle escaping off, you should escape any user input yourself!

Is this necessary in this case? And if so, how would we do that?

closes #728

### DIFF
--- a/src/renderer/components/file-manager/sources/webdav.js
+++ b/src/renderer/components/file-manager/sources/webdav.js
@@ -96,7 +96,8 @@ class Webdav extends PureComponent {
         console.error(err);
         showDialog(
           t('error.webdav-connection-failed-info', {
-            endpoint
+            endpoint,
+            interpolation: { escapeValue: false }
           })
         );
       });


perry-mitchell:
> Is this necessary in this case? And if so, how would we do that?

I don't personally think it's necessary here. We _could_ do URL validation but that seems like overkill really.

PR #731, opened 2019-01-04
Enhancement: Add a tool-tip for the password generator

### LINKED ISSUE #713: As a user, when I hover buttons, I want (title?) tooltips on everything so I know what buttons do #Enhancement
Hi all,

Hope you're well

Thanks for software

As a user, when I hover buttons (notably in edit state), I want (title?) tooltips on everything so I know what they do

Screen shot/cast:

foo mov

If you're reading this and:

a) agree, then please up-vote and/or reply +1, or;
b) disagree, then I'd love to know why?

Unless you specify a time frame, I'll follow up again in the near future and base my decision to keep using buttercup on whether this is fixed

If you have any issues (e.g. questions/queries) happy to help

Hope this helps!

Keep up great work!

Cheers

derek-li:
Hi guys!

This pull request adds a tool tip for the password generator feature. I've added translations for the supported languages as well. Please let me know if you'd like to see any changes!

Fixes #713 

Tool tip:
screenshot1

Language support:
screenshot2

### DIFF
--- a/locales/cs/base.json
+++ b/locales/cs/base.json
@@ -199,6 +199,9 @@
     "confirm": "Potvrdit",
     "nevermind": "Opustit"
   },
+  "password-generator": {
+    "password-generator": "Generovat heslo"
+  },
   "search": {
     "search": "Hledat"
   },
--- a/locales/de/base.json
+++ b/locales/de/base.json
@@ -201,6 +201,9 @@
     "nevermind": "Abbrechen",
     "new-password": "Neues Passwort"
   },
+  "password-generator": {
+    "password-generator": "Generiere Passwort"
+  },
   "search": {
     "search": "Suche"
   },
--- a/locales/en/base.json
+++ b/locales/en/base.json
@@ -217,6 +217,9 @@
     "confirm": "Confirm",
     "nevermind": "Nevermind"
   },
+  "password-generator": {
+    "password-generator": "Generate Password"
+  },
   "search": {
     "search": "Search"
   },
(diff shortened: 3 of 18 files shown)


perry-mitchell:
Any chance we could grab a screenshot from you before we dive in? 😊

derek-li:
@perry-mitchell 

Thanks for the quick response! I've added the requested screenshots.

I also was thinking about the original issue (password generator unclear) and am currently considering another solution by removing the magicIcon completely and instead explicitly state in text that the button is a password generator, similar to the save/cancel buttons. I think it'd be more clear. What do you think?

perry-mitchell:
> I've added the requested screenshots. 

Thanks! I've passed this to @sallar to review.

> considering another solution by removing the magicIcon completely and instead explicitly state in text that the button is a password generator

Hmm, not sure that I agree with that, as I feel that a good icon can definitely communicate what it's doing. It also doesn't require translation (to that effect, having less text and more iconography can definitely be a good thing). Also, space is somewhat limited:

image

I would suggest finding a more descriptive icon if you think it can be explained better with another image, but I don't personally think that text would work better in that case.

sallar:
Thanks @derek-li, however this is breaking the tests since it's adding one extra button and the button count is now off. Could you please have a look at the tests as well?

derek-li:
@sallar I've fixed the button count. 

I'm unsure why the appveyor CI is still failing while the travis CI passes. I think this could be an issue with using arbitrary wait times between function calls in the test specifications. Would you mind sharing your thoughts?

Thanks!

sallar:
@derek-li Im re-running Appveyor to see whatsup

PR #738, opened 2019-01-25
Feature: Warn before losing unsaved entry (edit mode)

### LINKED ISSUE #735: prevent loss of unsaved entry
when editing an existing entry, I often find myself forgetting to actually save the changes, and later switching to another entry to copy a different password. this way i lose the changes in the unsaved entry which i was editing.
a possible solution to avoid this problem would be a dialog that appears when trying to open another entry while the current one isn't saved, saying "you didn't save your entry, you will lose the data if you don't save it now" or simply not allowing the user to switch to a different entry.

also, you should somehow prevent closing the buttercup window (app) while adding a new entry or editing an existing entry, or at least ask the user if they are sure they want to quit without saving changes.

issue is similar to #634 but refers to entries that already existed and are being edited (not new entries)

wolffles:
Fixes #735 Now when clicking somewhere else during edit mode it will warn you of losing unsaved work. 

notes: I refactored some of the previous code because some of it seemed unnecessary like the parameter `isSavingNewEntry` which is always false and no code toggles it. I didn't understand the currentEntry check; I think checking the currentEntryMode will suffice.

### DIFF
--- a/src/shared/actions/entries.js
+++ b/src/shared/actions/entries.js
@@ -6,7 +6,6 @@ import { getQueue } from '../../renderer/system/queue';
 import {
   getCurrentGroupId,
   getCurrentArchiveId,
-  getCurrentEntry,
   getCurrentEntryMode,
   getExpandedKeys
 } from '../selectors';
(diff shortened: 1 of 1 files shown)


wolffles:
oh I see, I didn't try doing that when i was testing it. I'll take another look.

wolffles:
@ph1p, I added the isSavingNewEntry back and it solves this save issue, It seems though being able to save a label without a value is suppose to be how it is so I didn't adjust it.

I wonder if i was wrong about the currentEntry variable as well?

PR #741, opened 2019-01-26
added isSavingNewEntry again to prevent a bug

wolffles:
in response to https://github.com/buttercup/buttercup-desktop/pull/738

As @ph1p stated there was a bug when saving incomplete entries, when the parameter isSavingNewEntry was a necessary piece of the code. I added it again. My bad :[.

### DIFF
--- a/src/shared/actions/entries.js
+++ b/src/shared/actions/entries.js
@@ -25,10 +25,14 @@ import { setExpandedKeys } from '../../shared/actions/ui';
 import { loadOrUnlockArchive } from '../../shared/actions/archives';
 import { loadGroup } from '../../shared/actions/groups';
 
-export const selectEntry = entryId => async (dispatch, getState) => {
+export const selectEntry = (entryId, isSavingNewEntry = false) => async (
+  dispatch,
+  getState
+) => {
   try {
     const currentEntryMode = getCurrentEntryMode(getState());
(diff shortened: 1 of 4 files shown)


ph1p:
Hey! (: I think there is still an mistake and it was there already before your change ... I've found this bug by clicking around while editing or creating an entry. 
I think it's not your fault and I've created this tiny bug a while ago.

wolffles:
yeah I noticed it to but this pull request is just to fix the immediate issue that was merged

src/shared/actions/entries.js
```diff
@@ -25,12 +26,17 @@ import { setExpandedKeys } from '../../shared/actions/ui';
...
+) => {
   try {
+    const currentEntry = getCurrentEntry(getState());
     const currentEntryMode = getCurrentEntryMode(getState());
-    currentEntryMode === 'new' || currentEntryMode === 'edit'
+    (currentEntryMode === 'new' && !isSavingNewEntry) ||
+    (currentEntryMode === 'edit' && !isSavingNewEntry)
       ? showConfirmDialog(
-          i18n.t('entry.quit-unsave-entry'),
+          i18n.t('entry.quit-unsave-entry', console.log(currentEntry)),
```
perry-mitchell: Should this console log be removed?

Issue #734, opened 2019-01-18
Exception when starting Windows App


ajfoster85:
When starting the Windows app, an error screen appears with the following exception details:

> Unhandled Error
Error: listen EADDRINUSE :::12821
    at Function.y.listen (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:321:4541)
    at createHost (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:199:240237)
    at startFileHost (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:171:26943)
    at fc (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:406:123939)
    at C:/Program Files/Buttercup/resources/app.asar/dist/app.js:406:130490
    at Nc (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:406:130495)
    at App.i.app.on (C:/Program Files/Buttercup/resources/app.asar/dist/app.js:406:134595)

Version is 1.14.0 and it's a fresh install (I've only just started using it). I am also using the Firefox extension on the same computer (v2.3.0).

perry-mitchell:
It seems that something's listening on that port already, which is weird. It's not reserved in the IANA port registrations, so I can't imagine that another process would be randomly using it.

Would it be possible for you to [check what process is using that port](https://stackoverflow.com/questions/48198/how-can-you-find-out-which-process-is-listening-on-a-port-on-windows)?

AndreasFurster:
I'm not sure if the referenced commit solved the whole problem. I am experiencing the same issue and have some more info. 

I think the issue occurs when Buttercup is initially started by the browser extension and then opened in Windows. 

After opening a couple of instances via the browser or Windows I have a lot of instances of Buttercup. Even after closing all windows the processes stay active. At the moment I have 7 processes of Buttercup without a opened window. There is however only one system-tray icon.

I have not looked into the internals of Buttercup, but I suspect there is a http listener started multiple times. This causes to have the port used multiple times by Buttercup.

perry-mitchell:
Thanks for checking on this! It could be that the listener/server is not closed properly on shutdown, or something to that effect.

perry-mitchell:
Closing as the issue relates to a port in use. Either close the other app using the port, or disable Buttercup's browser connection feature.

PR #754, opened 2019-03-07
Added bugfix for Issue #734

rwolf449:
If you close the app on Windows right now, there are several processes still running.
Buttercup does not close properly which leads to the behaviour described in the issue.
I would suggest this small change, Buttercup now closes as intended (tested on Windows and Mac).

### DIFF
--- a/src/main/app.js
+++ b/src/main/app.js
@@ -26,7 +26,6 @@ const storage = pify(jsonStorage);
 const windowManager = getWindowManager();
 
 let appIsReady = false;
-let appTriedToQuit = false;
 let initialFile = null;
 
 // Crash reporter for alpha and beta releases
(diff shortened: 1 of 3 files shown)


src/main/app.js
```diff
@@ -147,10 +145,7 @@ app.on('ready', async () => {
   // When user closes all windows
   // On Windows, the command practice is to quit the app.
   app.on('window-all-closed', () => {
-    if (
-      appTriedToQuit ||
```
sallar: Why is this deleted? this is important for saving unsaved data. Check the rest of the file for this `appTriedToQuit`

rwolf449: I added appTriedToQuit again, and it still works, the checks also passed now. The problem seems to be `!getSetting(store.getState(), 'isTrayIconEnabled'`

sallar: @rwolf449 Then we have to figure out a way to fix it, not remove it. Since all OS's rely on this setting to NOT quit if the tray icon is enabled.

sallar:
Hey @rwolf449 any updates on this by any chance?

perry-mitchell:
@sallar Should we close this?

sallar:
Yes @perry-mitchell unfortunately this is not a solution to the problem. Apologies @rwolf449

PR #762, opened 2019-04-12
Issue626:  Users password is now cleared from the clipboard after a short duration

jadach1:
Before:
Password was being cleared for clipboard when user would enter ctrl-c, but not when the user pressed the button right of the password.

Now:
Password is cleared from the clipboard for both scenarios.

### DIFF
--- a/src/renderer/system/utils.js
+++ b/src/renderer/system/utils.js
@@ -8,16 +8,20 @@ const __cache = {
 
 const currentWindow = remote.getCurrentWindow();
 
-export function copyToClipboard(text) {
+export function copyToClipboard(text, isPassword) {
   clipboard.writeText(text);
 
-  // Clean the clipboard after 15s
-  // @TODO: Make a UI for this.
-  __cache.timer = setTimeout(function clipboardPurgerClosure() {
-    if (readClipboard() === text) {
-      copyToClipboard('');
+  // isPassword is a boolean
+  if (isPassword) {
+    if (__cache.timer) {
+      clearTimeout(__cache.timer);
     }
-  }, ms('15s'));
+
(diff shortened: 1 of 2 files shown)


src/renderer/system/utils.js
```diff
@@ -1,10 +1,23 @@
...
+import ms from 'ms';
+
+const __cache = {
+  timer: null
+};
 
 const currentWindow = remote.getCurrentWindow();
 
 export function copyToClipboard(text) {
   clipboard.writeText(text);
```
sallar: You need to clear the cache here. Similar to what happens in the other file?
https://github.com/buttercup/buttercup-desktop/blob/9f7fd929ae5c14986a12775f646d8a2135081797/src/renderer/system/shortcuts.js#L16

jadach1: Hey, thank you for the reply!  
I copied the function from shortcut.js to utils.js exactly, so I am not quite sure as to what you are referring to in regards to clearing the cache similar to the other file?  Thanks !

sallar: Hi! If you check `shortcut.js` there is a few more lines in the beginning of the function which you need to copy as well

jadach1: Thanks, I will take a look.

jadach1: So I added in a bit more in addition to clearing the cache.  Basically what I overlooked before was that the method is being called by every "copy-button" not just the password and this caused the clipboard to be cleared after 15 seconds regardless of whether it was a password or not.

This new change basically ensures that only the password will be cleared from the clipboard after 15 seconds.

I hope this helps 

Let me know your thoughts on the matter , thanks !

sallar:
Thanks!

Issue #717, opened 2018-11-30
Show a warning when saving two custom fields with the same label


uwqwxing886:
If you save an entry after adding two custom fields with the same label (like "URL") only the last one will be saved.

uwqwxing886:
I'm new to the code but it seems like the underlying structure is a key-value store that doesn't allow multiple fields with the same key (label), therefore that is why a warning message was suggested.

perry-mitchell:
Good idea! I would also prevent saving for such situations.

ldexterldesign:
I lost some crucial banking/finance data today because of this issue. It's not the first time it's happened (usually duplicate data which I tolerate) but it is the first time it's been important data.

Please consider higher priority or even bug status

Yours hopefully

perry-mitchell:
@ldexterldesign Agreed it could be higher prio. We don't currently have bandwidth for it, and would hope that someone could contribute it. Sorry that you lost some data.

keywordnew:
I hit this bug today. Saved 10 custom fields (as a workaround for not being able to create a note-like item instead of a login entry/item) and it only kept the last one.

PR #763, opened 2019-04-13
Added validation for duplicate labels

AlvinValdez:
Fixes issue #717

Before:
If you save an entry after adding two custom fields with the same label (like "URL") only the last one will be saved. No warning is displayed.

Now:
Warning is now displayed before you save entry.

### DIFF
--- a/src/shared/buttercup/entries.js
+++ b/src/shared/buttercup/entries.js
@@ -57,6 +57,18 @@ export function validateEntry(entry) {
     if (fields.filter(field => !field.property).length > 0) {
       errorMessages.push(i18n.t('entry.custom-fields-label-empty-info'));
     }
+
+    if (fields.length > 3) {
+      var titleofFields = [];
+
+      for (var i = 3; i < fields.length; i++) {
+        titleofFields.push(fields[i].property);
+      }
+
+      if (titleofFields.length !== new Set(titleofFields).size) {
+        errorMessages.push(i18n.t('entry.duplicate-label-found'));
+      }
+    }
   }
 
   if (errorMessages.length > 0) {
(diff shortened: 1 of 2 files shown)


src/shared/buttercup/entries.js
```diff
@@ -57,6 +57,18 @@ export function validateEntry(entry) {
     if (fields.filter(field => !field.property).length > 0) {
       errorMessages.push(i18n.t('entry.custom-fields-label-empty-info'));
     }
+
+    if (fields.length > 3) {
+      var titleofFields = [];
```
perry-mitchell: Please use `const`/`let` instead to prevent hoisting.

src/shared/buttercup/entries.js
```diff
@@ -57,6 +57,18 @@ export function validateEntry(entry) {
     if (fields.filter(field => !field.property).length > 0) {
       errorMessages.push(i18n.t('entry.custom-fields-label-empty-info'));
     }
+
+    if (fields.length > 3) {
```
sallar: number of "default fields" are not always 3. 
You can find the custom fields like so:

```js
const removeableFields = fields.filter(field => field.removeable);
if (removeableFields.length > 0) {
```

src/shared/buttercup/entries.js
```diff
@@ -57,6 +57,18 @@ export function validateEntry(entry) {
...
+      var titleofFields = [];
+
+      for (var i = 3; i < fields.length; i++) {
```
sallar: instead of this forEach you can do so:

```
removeableFields.forEach(field => {
```

sallar:
Hey @AlvinValdez, any chance you could look at the requested changes? :) Thank you!

sallar:
Since there is no answer, closing in favor of #813

PR #800, opened 2019-08-09
Password clipboard behavior changes

### LINKED ISSUE #626: Copied password remains in clipboard after clearing-time
Hi,
I noticed that the copied passwords remain in the clipboard.
Also in the version windows 10 17704 (insider) there is a history of the clipboard, and we find the password in clear. 

do you think that it will be possible to correct this ?

image

### LINKED ISSUE #774: Password field: select and copy issue
When you select the password field (hold down mouse or triple click), and copy this field, it actually copy "●●●●●●●●●●", and not the expected password.   
You need to use the dedicated (and hard to reach buton to reveal it).

copy_password  

I would love to be able to copy this field like any other.

When login to a page, I first copy the URL, then the username, finally the password.
Having different way to copy these information just get in the way and bug me.

Thanks for this awesome software! Keep up the good work

Application: Buttercup v1.15.0
Node.js: v8.11.4
System: Ubuntu Linux 4.18.0-21-generic #22-Ubuntu SMP Wed May 15 13:13:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

lotarwalace:
Two small changes around the clipboard behavior for passwords.

- 15s password clearing only applied if clipboard has not changed
- Concealed password can be copied directly

I think these changes should satisfy #774. The use of `clear` might even resolve #626 (though I don't have a windows machine to test this with)

Not sure about the use of `!important` on the style...
Let me know what you guys think!

Fixes #774
Fixes #626

### DIFF
--- a/src/renderer/components/archive/copyable.js
+++ b/src/renderer/components/archive/copyable.js
@@ -14,6 +14,7 @@ const Password = styled(ColoredDigits)`
   font-family: Anonymous;
   font-size: 14px;
   font-weight: bold;
+  user-select: all !important;
 
   .num {
     color: var(--brand-primary-darker);
@@ -106,6 +107,10 @@ class Copyable extends PureComponent {
         role="content"
         value={content}
         concealed={this.state.concealed}
+        onCopy={e => {
+          e.preventDefault();
+          this.handleCopy(true);
+        }}
       />
     );
   }
--- a/src/renderer/system/shortcuts.js
+++ b/src/renderer/system/shortcuts.js
@@ -1,22 +1,14 @@
 import Mousetrap from 'mousetrap';
-import ms from 'ms';
 import { lockArchive } from '../../shared/actions/archives';
 import { getCurrentArchiveId, getCurrentEntry } from '../../shared/selectors';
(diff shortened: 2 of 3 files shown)


src/renderer/components/archive/copyable.js
```diff
@@ -14,6 +14,7 @@ const Password = styled(ColoredDigits)`
   font-family: Anonymous;
   font-size: 14px;
   font-weight: bold;
+  user-select: all !important;
```
perry-mitchell: I don't think `!important` is required as styled-components sets it with an exact class, as it does with all other components. This value should override any inherited style.

lotarwalace: Hmmm, I looked into why I had to do that the other day, and it looks like the `role="content"` attribute selector  in `global-styles.js` ends up being more/as specific as the styled-components class selector.

I guess I could do one of the following:
 - just forget about the selection behavior and not add the `user-select` style on the styled-component class selector
- Leave the `!important`
- Remove the `role` attribute on the password component

perry-mitchell: I think the `!important` is fine in this case, imo.

PR #813, opened 2019-10-06
fix: display an error when an entry has duplicate custom fields

antoine-amara:
when a user creates custom fields with the same label, an error
message has to be throw by the form validator. if not, the custom
field is erased and only the last one is saved.

update the entry form validation function to be able to return a form
error, when we have a duplicate custom field, declared in an entry.
this action invalid the entry and prevent an eventual data loss by
canceling the update action.

Fixes: https://github.com/buttercup/buttercup-desktop/issues/717


sallar:
Im merging this to the other branch to test why the test fails @perry-mitchell @antoine-amara

PR #816, opened 2019-10-12
feat: adds the ability to show/hide the dock icon (MacOS only)

### LINKED ISSUE #806: Adding option to remove Dock Icon
This strictly has to do with **macOS**, I'm not sure how this would affect any other OS. I'm not sure if this is possible from Electron (from my understanding that's what the desktop client is built with), but I believe it is. Is it possible to add a setting to remove the dock icon and and the app window from always being present? Kind of like having the app run in the background with just the menu bar icon. A few apps already do this on mac (with a setting). NordVPN being one, OneDrive (and other cloud syncing apps), etc. 

NordVPN example (the app is running in the background with no open window or dock icon):
image

EDIT: This actually already happens if you open Buttercup (have the setting to enable the menubar icon enabled) and then close the window. Clicking the menu bar icon opens the window app again and shows the dock icon which is a good enough amount of functionality I think. It would maybe just be nice to have a setting to set this as a default. I want to set Buttercup to start on startup. But, I don't care for the window popup as I don't always need the app every minute of the day. So just leaving it in the background is ideal.

A setting in this menu would be where it would be implemented I imagine.
image

gillesdemey:
fixes #806

I still need to go through the translations, but the functionality is ready to be reviewed :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,11 @@
     "nsis": {
       "perMachine": true
     },
+    "mac": {
+      "extendInfo": {
+				"LSUIElement": 1
+			}
+    },
     "win": {
       "target": "nsis"
     },
--- a/package.json
+++ b/package.json
@@ -42,7 +42,10 @@
     "appId": "pw.buttercup.desktop",
     "copyright": "Copyright © 2015-2018 MadDev Oy",
     "mac": {
-      "category": "public.app-category.productivity"
+      "category": "public.app-category.productivity",
+      "extendInfo": {
+        "LSUIElement": 1
+      }
     },
     "dmg": {
       "background": "./build/background@2x.png",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -67,6 +67,11 @@
     "nsis": {
       "perMachine": true
     },
+    "mac": {
+      "extendInfo": {
+				"LSUIElement": 1
```
sallar: Can you change the tabs used here to spaces?

gillesdemey: done :)

perry-mitchell:
Closing this as we're preparing a v2 merge _very_ soon.. This PR doesn't align with the new codebase nor utilities used within it. Once V2 has been merged it'd be great to reconsider such a feature in a refreshed format, if needed. Thank you!

PR #841, opened 2019-11-04
Added build.win.publisherName in package.json to (hopefully) fix…

### LINKED ISSUE #805: Windows 10: update procedure does not work
### Problem
The update to 1.16.2 failed on my Windows 10 PC. After some time, I got an "EACCESS" error or something like that. I can't reproduce it, because I already found a workaround: If buttercup is run as admin, the update works.

### Suggestion:
I think buttercup should be run as administrator permanently. Could you implement it so that this issue is fixed for all windows users?

ckadluba:
According to comments in a similar issue in the electron repo (https://github.com/electron-userland/electron-builder/issues/3667#issuecomment-463227859) adding the build.win.publisherName in package.json could fix issue #805. I could not verify this though, because I cannot easily test the update mechanism.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
       "perMachine": true
     },
     "win": {
-      "target": "nsis"
+      "target": "nsis",
+      "publisherName": "MadDev Oy"
     },
     "linux": {
       "target": [


sallar:
Thanks @ckadluba Im going to try to re-release.

Issue #856, opened 2019-11-29
Fix sidebar background colour #Bug


ldexterldesign:
<img width="1440" alt="Screenshot 2019-11-29 at 22 42 17" src="https://user-images.githubusercontent.com/1715809/69891029-c3d69c80-12f9-11ea-83ca-255390d1e227.png">

ldexterldesign:
First BC PR 😟

To reproduce this issue launch BC and then cmd + r to refresh. Because cmd + r, I assume it's some kind of dynamic props issue but I've attempted a fix by simplifying the CSS...

I don't see the need to use conditionals `isOSX()` in the CSS but if there is a need then I'd love to know why? Also, hsl(a) > rgb(a) > hex if one fancies giving global-styles.js some love.

NB please see my TODO in the PR as I'd hate this to break something I can't test

Hope this helps

Regards 😟

sallar:
Thanks @ldexterldesign. We need the `isOSX()` calls, because the app looks different in macOS and there is vibrancy and translucency.

![image](https://user-images.githubusercontent.com/768052/69892422-f4c0cc80-130d-11ea-8c91-79e6a20ecd19.png)

ldexterldesign:
Why's this platform (i.e. macOS) specific?

Cheers

sallar:
@ldexterldesign Windows and Linux don't support vibrancy.

ldexterldesign:
https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/translucency/

I see

If you can reproduce this issue then let me know and I'll take another look over the weekend

Cheers

ldexterldesign:
BUMP

ldexterldesign:
BUMP

> Can you reproduce this?

Cheers

sallar:
The issue can be reproduced by opening and closing developer tools (`cmd+option+I`). I'm just not sure what can be done about it.

ldexterldesign:
Will take another look at he PR ASAP

Regards

perry-mitchell:
I wasn't seeing this issue in development on 1.18.1 / 1.18.2:

![image](https://user-images.githubusercontent.com/3869469/72187404-390cb580-3400-11ea-8a76-3f900fb71df7.png)

Anyone else seeing it?

ldexterldesign:
I just experienced this issue in https://github.com/Sonarr/Sonarr/issues/3218 so I presume it's something in a core library, not buttercup-specific

Hope this helps

[...]

PR #857, opened 2019-11-29
Fix sidebar background colour #856

ldexterldesign:


ldexterldesign:
https://github.com/buttercup/buttercup-desktop/issues/856#issuecomment-565060784

perry-mitchell:
Not sure what the status of this is? Seems to be stale and I don't think the query @sallar had has been resolved. I'll close this for now. We can reopen if we find that it is indeed still valid.

ldexterldesign:
Correct to close this but let's leave https://github.com/buttercup/buttercup-desktop/issues/856 open as it's still a bug I can reproduce (and I've moved to a new MacBook Pro since opening the issue)

Apologies/regards

PR #867, opened 2019-12-14
Adding why-did-you-render

Rafi993:
This PR adds why-did-you-render to detect unwanted re-renders in the app

### DIFF
--- a/src/shared/store/configure-store.js
+++ b/src/shared/store/configure-store.js
@@ -1,7 +1,10 @@
+import React from 'react';
 import { createStore, applyMiddleware, compose } from 'redux';
 import thunk from 'redux-thunk';
 import { createLogger } from 'redux-logger';
 import { electronEnhancer } from 'redux-electron-store';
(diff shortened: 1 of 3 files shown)


src/shared/store/configure-store.js
```diff
@@ -13,6 +16,13 @@ export default function configureStore(initialState, scope = 'main') {
   let middleware = [thunk];
 
   if (scope === 'renderer' && process.env.NODE_ENV === 'development') {
+    whyDidYouRender(React, {
+      include: [/^Connect/],
+      logOnDifferentValues: true,
+      trackHooks: true,
+      collapseGroups: true
+    });
```
perry-mitchell: Hmm, I wonder if this should only be enabled via a flag. We probably don't want it running 100% of the time - Probably only when we want to do specific debugging relating to multiple component renders. @sallar?

perry-mitchell:
Closing due to inactvity, and the fact that we're releasing a new build soon: #974

PR #875, opened 2019-12-28
Fix google drive auth

### LINKED ISSUE #850: Can't connect Google Drive
If i try to connect to Google Drive i get this error in version 1.17.2

In the Log there is no hint.

buttercup

### LINKED ISSUE #866: No version in help menu
MacOS Catalina:
image

sallar:
Fixes #850
Fixes #866

### DIFF
--- a/src/main/actions.js
+++ b/src/main/actions.js
@@ -1,5 +1,7 @@
 import debounce from 'lodash/debounce';
-import { ipcMain as ipc, BrowserWindow, app } from 'electron';
+import http from 'http';
+import { ipcMain as ipc, BrowserWindow, app, shell } from 'electron';
+import { ipcMain as betterIpc } from 'electron-better-ipc';
 import { getWindowManager } from './lib/window-manager';
 import { openFile, newFile, openFileForImporting } from './lib/files';
 import { setupMenu } from './menu';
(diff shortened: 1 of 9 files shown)


src/main/actions.js
```diff
@@ -62,3 +63,39 @@ export function setupActions(store) {
...
+  }
+}
+
+export function handleProtocolCall(url) {
+  if (typeof url !== 'string' || url.length === 0) {
+    log.warn('Empty protocol call. Abborting.');
+    return;
+  }
+  const path = url.replace('buttercup://', '');
+  const parts = path.split('/');
```
perry-mitchell: Could this be destructured instead, like say `const [ action ] = path.split('/');` ?

src/main/app.js
```diff
@@ -84,10 +84,20 @@ if (!lock) {
   app.quit();
 }
 
-app.on('second-instance', () => {
-  const focusedWindow = getMainWindow();
-  if (!focusedWindow) {
-    windowManager.buildWindowOfType('main');
+app.on('second-instance', (event, args) => {
+  reopenMainWindow(() => {
```
perry-mitchell: Noticed that you don't check the `getMainWindow()` result now - what does `reopenMainWindow` do if there's already another app running? I checked the code briefly not not sure as to how to read it exactly. Does it simply ignore the other instance and open anyway?

src/shared/buttercup/entries.js
```diff
@@ -57,6 +57,12 @@ export function validateEntry(entry) {
     if (fields.filter(field => !field.property).length > 0) {
       errorMessages.push(i18n.t('entry.custom-fields-label-empty-info'));
     }
+
+    const fieldKeys = fields.map(field => field.property);
```
perry-mitchell: This should maybe be (due to attributes):

```javascript
const fieldKeys = fields.reduce((output, field) => {
	if (field.propertyType === "property") {
		output.push(field.property);
	}
	return output;
}, []);
```

Otherwise if an attribute of the same name as a property exists, this will fail.

src/main/app.js
sallar: @perry-mitchell yeah this second instance thing is a bit tricky to read in general. So how this works is, if one instance is open, second instance simply doesn't open at all (it quits) and this `second-instance` event is fired in the *first* instance. So when this even is triggered, we simply bring back the main window if it's out of focus

sallar: @perry-mitchell yeah this second instance thing is a bit tricky to read in general. So how this works is, if one instance is open, second instance simply doesn't open at all (it quits) and this `second-instance` event is fired in the *first* instance. So when this even is triggered, we simply bring back the main window if it's out of focus

src/main/actions.js
sallar: @perry-mitchell I could do this: `const [action, ...args] = ` because I need the args a few lines below

sallar:
@perry-mitchell changes done :+1:

PR #883, opened 2020-01-19
Preferences Window - Changes

ph1p:


src/renderer/components/preferences/general.js
```diff
@@ -140,7 +144,9 @@ class General extends PureComponent {
           type="number"
           min="0"
           defaultValue={DEFAULT_CLIPBOARD_CLEAR_SECONDS}
-          onChange={e => setSecondsUntilClearClipboard(e.target.value)}
+          onChange={e =>
+            setSecondsUntilClearClipboard(e.target.value, 'number')
```
perry-mitchell: Why not parse this before sending it in, rather than adding a param to the function? :)

src/renderer/components/preferences/general.js
```diff
@@ -149,7 +155,7 @@ class General extends PureComponent {
           type="number"
           min="0"
           defaultValue={DEFAULT_ARCHIVE_CLOSE_SECONDS}
-          onChange={e => setAutolockSeconds(e.target.value)}
+          onChange={e => setAutolockSeconds(e.target.value, 'number')}
```
perry-mitchell: Same question as earlier - perhaps easier to parse before setting?

PR #884, opened 2020-01-28
Allow import passwords from browser csv dumps

antonvolokha:
Fixed closing the window during the creation of a new archive when file extension is not specified.
Also added password import from CSV browser dump.

Checked on Ubuntu 19.10

### DIFF
--- a/src/main/lib/buttercup.js
+++ b/src/main/lib/buttercup.js
@@ -3,7 +3,8 @@ import {
   importFrom1PIF,
   importFromLastPass,
   importFromBitwarden,
-  importFromButtercup
+  importFromButtercup,
+  importFromCSV
 } from '@buttercup/importer';
 import { ImportTypes } from '../../shared/buttercup/types';
 
@@ -33,6 +34,8 @@ export function importArchive(type, filename, password) {
       return importFromButtercup(filename).then(archive =>
         archive.getHistory()
       );
+    case ImportTypes.BROWSERS:
+      return importFromCSV(filename).then(archive => archive.getHistory());
     default:
       throw new Error('Wrong import type provided');
   }
--- a/src/main/lib/buttercup.js
+++ b/src/main/lib/buttercup.js
@@ -34,7 +34,7 @@ export function importArchive(type, filename, password) {
       return importFromButtercup(filename).then(archive =>
         archive.getHistory()
       );
-    case ImportTypes.BROWSERS:
+    case ImportTypes.CSV_GENERIC:
       return importFromCSV(filename).then(archive => archive.getHistory());
     default:
       throw new Error('Wrong import type provided');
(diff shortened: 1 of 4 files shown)


src/main/lib/buttercup.js
```diff
@@ -33,6 +34,8 @@ export function importArchive(type, filename, password) {
       return importFromButtercup(filename).then(archive =>
         archive.getHistory()
       );
+    case ImportTypes.BROWSERS:
```
perry-mitchell: Would be good to use `CSV_GENERIC` at the import type.

src/shared/buttercup/types.js
```diff
@@ -12,6 +12,7 @@ export const ImportTypes = {
   KEEPASS: 'keepass',
   LASTPASS: 'lastpass',
   ONE_PASSWORD: '1password',
+  BROWSERS: 'browsers',
```
perry-mitchell: Probably best to use `CSV_GENERIC: 'csv_generic'` as this isn't just related to browsers :)

src/shared/buttercup/types.js
```diff
@@ -40,6 +41,12 @@ export const ImportTypeInfo = {
     password: false,
     name: 'Buttercup (CSV)',
     extension: 'csv'
+  },
+  [ImportTypes.BROWSERS]: {
```
perry-mitchell: Same change here

src/shared/buttercup/types.js
```diff
@@ -40,6 +41,12 @@ export const ImportTypeInfo = {
...
+    password: false,
+    name: 'Browsers (CSV)',
```
perry-mitchell: Should be `Generic (CSV)`

perry-mitchell:
This also needs the buttercup importer version bumped to 1.2.0

Badman12:
I renamed type to CSV_GENERIC in all places

perry-mitchell:
@Badman12 Any chance you could update to 1.2.0 of the importer package? Otherwise this functionality won't work.

Badman12:
@perry-mitchell Update in package.json, right? sorry i forgot about it :)

perry-mitchell:
Ah damn, I just realised why we haven't updated importer earlier. It's broken due to argon2 support. Something we need to investigate before this is merged. I'll mark it on my high prio list.

Badman12:
Hm.. I build importer localy, run tests and everything looks fine

perry-mitchell:
@Badman12 The importer does in most places, except on Windows. At least last time we tried to build it for the desktop application on windows it fails.

perry-mitchell:
Closing this as we're preparing a v2 merge _very_ soon.. This PR doesn't align with the new codebase nor utilities used within it. Once V2 has been merged it'd be great to reconsider such a feature in a refreshed format, if needed. Thank you!

Issue #919, opened 2020-06-22
WebDav Service Returns 404 While Endpoint Tailed With Slashes


kenmjlee:
Hi,

After upgrading to 1.19.0, I discovered that one of my webdav connections returned status 404 because WebDav Client was used instead of WebDev-FS while releasing 1.17.0. And this issue had been confirmed and fixed by Perry, the WebDav client's author, for the [Issue #185 of WebDav](https://github.com/perry-mitchell/webdav-client/issues/185), and I am using the same WebDav service as well as this issue's creator. 

Would you mind considering a plan to update this package in the next release? 

Thank you.

perry-mitchell:
This app uses `webdav`, not `webdav-fs`, btw. I perform maintenance on WebDAV-fs more rarely than I do on the WebDAV client project. You're right that issues sounding similar to yours has been solved. 

We're about to release a new version, with a brand new Buttercup core build, that uses the latest build of webdav. Once this is out, let's revisit this to see if it's helped. Sorry for the inconvenience.

kenmjlee:
Yep, I know that after the release v1.17.0 the buttercup have been using the webdav. So I mentioned this issue was occurred by webdev but not by buttercup. But anyway, it sounds good to hear there will be a brand new release. 

Thanks for your effort.

sallar:
Fixed via #942

PR #941, opened 2020-08-27
Fixing Issue #919: WebDav Returns 404 While Endpoint Tailed With Slashes

kenmjlee:
Fixing Issue #919: WebDav Returns 404 While Endpoint Tailed With Slashes

Update the v2.10.1 of webdav package which has fixed the relevant issue.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -13,7 +13,7 @@
     "@buttercup/importer": "^0.15.0",
     "buttercup": "^2.16.1",
     "conf": "^1.0.0",
-    "webdav": "^2.9.1",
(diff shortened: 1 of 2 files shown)


sallar:
Thanks. Can you also please update `app/package.json`? That's what is used when packaged.

kenmjlee:
Done. It was updated accordingly.

sallar:
@kenmjlee thanks. it seems like that the test are failing because `package-lock.json` is not updated. Can you also please run `npm install` in the root and `app` so they are also updated?

PR #942, opened 2020-08-27
Further submission with package-lock jsons for fixing #919

kenmjlee:
Sorry for making you feel inconvenient. It seems that my base is not from the latest buttercup upstream. So I create a new PR to ensure no conflicts.


sallar (approved):
Great! Thank you

PR #950, opened 2020-09-27
#768 - Make fields with URL value clickable

### LINKED ISSUE #768: Make links clickable
When adding a link into a custom field, it would be helpful to be able to click on it to visit. It's possible to click to copy but not click to visit.

**Suggested solution:** On hover, add an extra button with a chain icon to visit the link

montrueque:
This PR offers a possible solution to the #768 issue. It adds a button similar to the existing copy or reveal password buttons, as well as the contextual menu option.  It also adds translation in all available languages.

Fixes #768

### DIFF
--- a/src/renderer/system/utils.js
+++ b/src/renderer/system/utils.js
@@ -2,13 +2,12 @@ import path from 'path';
 import { clipboard, remote, shell } from 'electron';
 import ms from 'ms';
 
+const currentWindow = remote.getCurrentWindow();
 const isValidUrl = require('is-valid-http-url');
 
 const __cache = {
   timer: null
 };
-
-const currentWindow = remote.getCurrentWindow();
 
 export function copyToClipboard(
   text,
@@ -54,7 +53,6 @@ export function openUrl(url) {
   if (!/^http(s)?:\/\//i.test(url)) {
     url = `https://${url}`;
   }
-
   shell.openExternal(url);
 }
 
--- a/src/renderer/system/utils.js
+++ b/src/renderer/system/utils.js
@@ -2,6 +2,8 @@ import path from 'path';
 import { clipboard, remote, shell } from 'electron';
 import ms from 'ms';
 
+const isValidUrl = require('is-valid-http-url');
+
 const __cache = {
   timer: null
 };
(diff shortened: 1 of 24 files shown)


src/renderer/system/utils.js
```diff
@@ -44,10 +44,25 @@ export function readClipboard() {
...
 
+export function isUrl(url) {
+  var pattern = new RegExp(
+    '^(https?:\\/\\/)?' +
+      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' +
+      '((\\d{1,3}\\.){3}\\d{1,3}))' +
+      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' +
+      '(\\?[;&a-z\\d%_.~+=-]*)?' +
+      '(\\#[-a-z\\d_]*)?$',
+    'i'
```
perry-mitchell, line 55: This pattern doesn't appear to support non-ascii characters, and users may store non-ascii URLs in non-punycode format. Imo it would be best to use a combination of [is-url](https://www.npmjs.com/package/is-url) and [is-valid-domain](https://www.npmjs.com/package/is-valid-domain).

If a string passes `is-url` and doesn't have the protocol prefix, ie. `my-domain.com`, pass it to `is-valid-domain` for confirmation that it's a domain and not some random string. This checks the string against valid domain TLDs.

Being that the desktop app is installed and it's already quite large, I don't see the harm in adding these extra packages to allow for this functionality.

src/renderer/system/utils.js
```diff
@@ -44,10 +44,25 @@ export function readClipboard() {
...
+  );
+
+  return !!pattern.test(url);
+}
+
 export function openUrl(url) {
-  if (!/^http(s)?:\/\//i.test(url)) {
+  if (!isUrl(url)) {
```
perry-mitchell, line 62: I would suggest reverting this functionality, as it's checking for a protocol. Might be best to use `^\w+:\/\/` or something else, and if it's false, assume `https`. The protocol is required for opening URLs, and we should default to https for security.

perry-mitchell (changes requested):

src/renderer/system/utils.js
montrueque: Yes, you're right. I'll revert it.

src/renderer/system/utils.js
montrueque: I had seen some extra package which do these checks, but I thought it was a better idea to implement directly. 

I change this part as you suggest.

src/renderer/system/utils.js
montrueque: Fixed.

src/renderer/system/utils.js
montrueque: Fixed.

perry-mitchell (approved):
Looks good to me

sallar (approved):
Looks great to me

PR #963, opened 2020-11-29
Remove node.js version in TravisCI $NAME variable, cc #799

PeterDaveHello:
As the node version is not defined in `.travis.yml` anymore, should also remove it in the `$NAME` variable, in case the version was updated in the future but the name wasn't, which will be confusing.


perry-mitchell (approved):
Thanks!

Issue #955, opened 2020-10-17
Cannot create local vault file on Linux using desktop client


wtsiamruk:
The Local vault cannot be create on linux

OS: Ubuntu 20.04 (GNOME DE)
Latest releease of buttercup at the moment of writing.

Steps to Reproduce.

Install fresh buttercup
Open the app
Click on add vault, create new
Choose a folder to save the file, enter the filename, click save.

Expected result: file is saved and opened as a vault
Actual: Nothing happens.
The logs on ~/.config/../log.log are also clear.

priyansh-anand:
Same issue
OS: Pop OS 20.04 (GNOME DE)

priyansh-anand:
Solution: It is working if I add ".bcup" at the end of filename

wtsiamruk:
> Solution: It is working if I add ".bcup" at the end of filename

Ok, so the fix would be to attach that ".bcup" when parsing filename :+1:

ailtonbsj:
Same problem. I was trying with ".bcp" kkkk

♬ Electrify my heart-art-art-art-art-art-oh-ooh-ooh 🎶

perry-mitchell:
This will be fixed in the upcoming major version (v2), but we'd still accept a PR for this on v1.

perry-mitchell:
Closing this as stale.

PR #954, opened 2020-10-11
feat: force .bcup extension for new vault

antoine-amara:
when creating a new vault the user enter a name. Be sure to force the
new vault name with .bcup extension to be able to finalize the
archive creation (open the archive and create a password).

Fixes: https://github.com/buttercup/buttercup-desktop/issues/860

PR #970, opened 2021-02-05
Fix bug when creating a new vault in Linux without '.bcup' extension

mikeg1440:
## Issue
Nothing happens when Linux users are prompted to set a file name for a new vault if 
they enter a filename without the '.bcup' file extension.  No file is created and no notification is displayed, the file dialog box just closes with no indication that it failed to create the vault other than the fact there is no new vault.  

## Expected 
Either add the extension automatically or notify user that the file name must contain a '.bcup' extension.

### Steps to Reproduce
Attempt to create a new vault without specifying the '.bcup' extension in the filename on Linux.

### Notes
It looks like code from a previous fix in the loadFile function returns empty if the filePath does not have the '.bcup' extension causing this to occur.

## Changes Made
I added code in the showSaveDialog function to check if the filePath ends with '.bcup' and if not then append it the string. 

Fixes: #955

### DIFF
--- a/src/main/lib/files.js
+++ b/src/main/lib/files.js
@@ -60,7 +60,7 @@ function showSaveDialog(focusedWindow) {
     })
     .then(({ filePath }) => {
       if (typeof filePath === 'string' && filePath.length > 0) {
-        if (path.extname(filePath).toLowerCase() != '.bcup') {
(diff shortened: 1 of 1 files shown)


perry-mitchell:
I think this was handled by an earlier MR, #954

Issue #990, opened 2021-04-11
Buttercup crashes immediately on Windows


vikaspotluri123:
Congrats on 2.0! 🥳

I just installed 2.0.0, but the app keeps crashing. Here's the log from launching it in powershell:

```
[INF] 20:48:42: Application ready
[INF] 20:48:42: Application session started: 2021-04-11T20:48:42.237Z
[INF] 20:48:42: Logs location: (%localappdata%)\Buttercup-nodejs\Log\buttercup-desktop.log
[INF] 20:48:42: System locale detected: en-US
[INF] 20:48:42: Starting with language: en-US
[INF] 20:48:42: Updated search records for 0 vaults
[INF] 20:48:42: Updated search records for 0 vaults
[ERR] 20:48:42: TypeError: electron_1.systemPreferences.canPromptTouchID is not a function
at <anonymous>              resources/app.asar/build/main/services/biometrics.js:82  return electron_1.systemPreferences.canPromptTouchID();
at next
at                          resources/app.asar/build/main/services/biometrics.js:8   step((generator = generator.apply(thisArg, _arguments || [])).next());
at new Promise
at __awaiter                resources/app.asar/build/main/services/biometrics.js:4   return new (P || (P = Promise))(function (resolve, reject) {
at supportsBiometricUnlock  resources/app.asar/build/main/services/biometrics.js:81  return __awaiter(this, void 0, void 0, function* () {
at                          resources/app.asar/build/main/actions/appMenu.js:34      const biometricsSupported = yield biometrics_1.supportsBiometricUnlock();
at next
at fulfilled                resources/app.asar/build/main/actions/appMenu.js:5       function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reΓÇª
```

The log file mentioned in line 3 does not exist.

It seems to be caused by #988; It looks like [`canPromptTouchID`](https://www.electronjs.org/docs/api/system-preferences#systempreferencescanprompttouchid-macos) is available on macos only

Alternus:
The same error is also occurring on Manjaro Linux when attempting to run the .appimage

perry-mitchell:
Fixed by #994

PR #994, opened 2021-04-12
Fixed Biometrics Crash

### LINKED ISSUE #993: Buttercup 2.0.0 not opening on Windows 10
Buttercup was working fine, but after unistalling and installing the 2.0.0 for Windows x64 downloaded from Github, the program won't open. Not even a process is initialized in Task Manager.

Alternus:
Fixed the crash due to `TypeError: electron_1.systemPreferences.canPromptTouchID is not a function` mentioned in #990 and likely also the cause of #993. 
Currently tested this on Manjaro Linux.

EDIT: Fixes #993

### DIFF
--- a/source/main/services/biometrics.ts
+++ b/source/main/services/biometrics.ts
@@ -67,7 +67,11 @@ export async function sourceEnabledForBiometricUnlock(sourceID: VaultSourceID): 
 }
 
 export async function supportsBiometricUnlock(): Promise<boolean> {
-    return systemPreferences.canPromptTouchID();
+    try {
+        return systemPreferences.canPromptTouchID();
+    } catch (err) {
+        return false;
+    }
 }
 
 async function storePassword(sourceID: VaultSourceID, password: string): Promise<void> {


source/main/services/biometrics.ts
```diff
@@ -67,7 +67,11 @@ export async function sourceEnabledForBiometricUnlock(sourceID: VaultSourceID):
 }
 
 export async function supportsBiometricUnlock(): Promise<boolean> {
-    return systemPreferences.canPromptTouchID();
+    try {
+        return systemPreferences.canPromptTouchID();
+    } catch (err) {
+        return false;
+    }
```
vikaspotluri123, line 74:
```suggestion
	if (process.platform === 'darwin') {
        return systemPreferences.canPromptTouchID();
    }

    return false;
```

I don't think it really matters either way 😄

jy95: In [Electron docs](https://www.electronjs.org/docs/api/system-preferences#systempreferencescanprompttouchid-macos), it is written that `systemPreferences.canPromptTouchID()` is a MacOS-only feature. So it really don't matter ...

PR #1057, opened 2021-06-09
fix(1011): search not working due to org.freedesktop.secrets missing

### LINKED ISSUE #1011: Search bar does not appear when using the ctrl + f
Hi, i an using buttercup `2.2.0` and when i press `ctrl + f` the search bar does not appear. I tried version `1.20.5` and it works fine.

I am using linux debian with KDE.

Another thing is the copy buttons in an entry, in the previous v1 release they were all vertically on the same column. that made them easy to click. In the current version they are after the entry's width. (This was solved)

maciek134:
Fixes #1011.

Not sure if the error handling is correct, let me know if I should handle it differently (I figured a warning is enough).


perry-mitchell:
Thanks @maciek134 that's fantastic! Great that you found this.

perry-mitchell (approved):

PR #1236, opened 2023-08-23
Maximizing the desktop application window upon opening

quintushr:
Maximizing the desktop application window upon opening

### DIFF
--- a/source/main/services/config.ts
+++ b/source/main/services/config.ts
@@ -13,7 +13,8 @@ const DEFAULT_CONFIG = {
     windowHeight: 600,
     windowWidth: 800,
     windowX: null,
-    windowY: null
+    windowY: null,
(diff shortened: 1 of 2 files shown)


quintushr:
Hi @perry-mitchell , do you think that maximizing the window or not upon launch should be in the settings, or should it be the default setting ?

perry-mitchell:
Hi @quintushr - thanks for the PR. This should definitely be behind a setting and _not_ be the default.

quintushr:
I saved the maximize state of the window, whether it was maximized or not, and upon launch, it is initialized with the parameter.

Is it better like this ?

perry-mitchell:
Yep, this is great! Thanks.

perry-mitchell (approved):

PR #1242, opened 2023-10-14
I18n for japanese.

satler-git:
Add japanese.
The i18n for buttercup/ui was already done by another actor.


perry-mitchell:
This is amazing @satler-git! Seems like we already have it [translated in the UI](https://github.com/buttercup/ui/blob/master/src/i18n/translations/ja.json), which is great. I'll merge this today.

perry-mitchell (approved):

Issue #1247, opened 2023-11-06
[feature request] Allow per-user installation


Oxalin:
Hi,

Since the extension will rely on the desktop application in the near future, I've been testing the desktop application in our workplace context. Users are not allowed to make per-machine installation, otherwise it must be made by a system admin. This is something new from the last couple of months, since we are tightening our user accesses across every computer provided by our organization to be inline with our confidentiality policy. Prior to this policy change, users were allowed to install and update applications that were install system wide. Also, this was not a problem with the extension, since it was/is installed per browser profile.

That being said, users are allowed to install per-user applications for permitted applications. Buttercup would be one of them.

Since we all rely on Buttercup to share a common vault (through Google Drive), it would really be usefull to have this per-user installation feature. Not a deal breaker, since our team is relatively small and it will still be manageable without this feature.

All computers provided by the office are running Windows. Some users are allowed to use their own personal computer (Windows and Mac), but they are expected to be able to install and manage their own applications according to our policies.

perry-mitchell:
Hello! This sounds like a feature that'd be dependent on our installation system, built into Electron Builder / updater. Someone would have to research if such a feature is available there. If it isn't, it'll unfortunately be out of reach for Buttercup too.

Due to the amount of work required to maintain Buttercup, it's not feasible to switch to a more complex/custom build system that might allow such a feature, if it is indeed not possible in the current setup. I hope it is, however, as I do fundamentally agree with the premise here.

I don't have time currently to check this, but hopefully someone can check the docs at some stage to see what's possible here.

Oxalin:
I'll have a look at it.

Oxalin:
According to NsisOptions, there is a perMachine parameter:
"perMachine = false Boolean - Whether to show install mode installer page (choice per-machine or per-user) for assisted installer. Or whether installation always per all users (per-machine)."
https://www.electron.build/generated/nsisoptions

In the package.json file, we have:
"nsis": {
      "artifactName": "${productName}-${os}-${env.NODE_ARCHITECTURE}-${version}-installer.${ext}",
      "perMachine": true,

So, this would need to be set to "false".

I'll try to setup my computer to build Buttercup from source and test this.

Oxalin:
By the way, ${env.NODE_ARCHITECTURE} was undefined when packaging and I had to modify it to ${arch}. Are you using a container where you define NODE_ARCHITECTURE when building? Otherwise, looking under node > process doesn't list this variable.

PR #1251, opened 2023-11-23
Modify NSIS options to allow installing Buttercup per-user

Oxalin:
Since Buttercup Desktop will be needed with Chrome's extension (manifest v3), users that were able to install the extension on their local account may end up being unable to install Buttercup Desktop if they don't have administrative privileges.

Adding oneClick:false and changing perMachine:false will launch an installation prompt giving the user a choice to install Buttercup per-user or per-machine (system wide).

Fixes [issue 1247](https://github.com/buttercup/buttercup-desktop/issues/1247)

Tested locally without any visible regression.

### DIFF
--- a/source/main/actions/config.ts
+++ b/source/main/actions/config.ts
@@ -15,9 +15,18 @@ export async function handleConfigUpdate(preferences: Preferences) {
     const language = getLanguage(preferences, locale);
     logInfo(` - Language updated: ${language}`);
     await changeLanguage(language);
+    logInfo(
+        ` - Auto clear clipboard: ${
+            preferences.autoClearClipboard ? preferences.autoClearClipboard + "s" : "Off"
+        }`
+    );
+    logInfo(
+        ` - Lock vaults after: ${
+            preferences.lockVaultsAfterTime ? preferences.lockVaultsAfterTime + "s" : "Off"
+        }`
+    );
+    logInfo(` - Start in background: ${preferences.startInBackground ? "Enabled" : "Disabled"}`);
     logInfo(` - File host: ${preferences.fileHostEnabled ? "Enabled" : "Disabled"}`);
-    logInfo(` - Auto clear clipboard: ${preferences.autoClearClipboard}s`);
-    logInfo(` - Lock vaults after: ${preferences.lockVaultsAfterTime}s`);
     if (preferences.fileHostEnabled) {
         await startBrowserAPI();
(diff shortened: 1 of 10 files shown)


source/main/actions/config.ts
```diff
@@ -6,18 +6,39 @@ import { getLanguage } from "../../shared/library/i18n";
...
+        }`
+    );
+
+    logInfo(` - Start in background: ${preferences.startInBackground ? "Enabled" : "Disabled"}`);
+
+    logInfo(
+        ` - Start with session launch: ${preferences.startWithSession ? "Enabled" : "Disabled"}`
+    );
+    await setStartWithSession(preferences.startWithSession);
+
```
perry-mitchell, line 40: Would you mind removing the extra lines here?

source/main/services/config.ts
```diff
@@ -89,3 +90,34 @@ export async function setVaultSettings(
...
+    const storage = getConfigStorage();
+    const preferences = await storage.getValue("preferences");
+    if (typeof preferences === "undefined") {
+        return false;
+    }
+    return preferences.startInBackground;
+}
+
+export async function setStartWithSession(enable: boolean): Promise<void> {
+    var AutoLaunch = require("auto-launch");
```
perry-mitchell, line 104: Why the require here? Can this just be imported?

Also, var => const..

source/main/services/config.ts
```diff
@@ -89,3 +90,34 @@ export async function setVaultSettings(
...
+
+    var autoLauncher = new AutoLaunch({
```
perry-mitchell, line 106: var => const :)

perry-mitchell (changes requested):
Some small items. Otherwise looks great!

source/main/services/config.ts
Oxalin: This is fixed by https://github.com/Oxalin/buttercup-desktop/commit/fd7ff2e9ed0da0bfa34ea124ab1eb122383dd6c9

source/main/services/config.ts
Oxalin: Also fixed by https://github.com/Oxalin/buttercup-desktop/commit/fd7ff2e9ed0da0bfa34ea124ab1eb122383dd6c9

perry-mitchell (approved):

perry-mitchell:
This is great!

image

